PR #2, opened 2024-11-18
Initial components

VictoriaBeilsten-Edmands:


src/components/Navbar.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+      />
+    ),
+  },
+};
+
+export const Links: Story = {
+  args: {
+    children: (
+      <NavLinks>
+        <NavLink href="#" key="proposal">Proposal</NavLink>
```
akademy, line 41: Let's make these a bit more generic
```suggestion
        <NavLink href="#" key="first">Link One</NavLink>
```

src/components/Navbar.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <NavLink href="#" key="visits">Visits</NavLink>
```
akademy, line 42:
```suggestion
        <NavLink href="#" key="second">Second Link</NavLink>
```

src/themes/DiamondTheme.ts
```diff
@@ -1,47 +1,86 @@
...
+          contrastText: "#000000", // black
+        },
+      },
+    },
+  },
+  components: {
+    MuiButton: {
+      styleOverrides: {
+        root: ({ theme }) => ({
+          borderBottom: "2px solid rgba(0,0,0,0.2)",
```
akademy, line 49: I don't think this is correct, as it's setting the bottom line of all the different variants of the button to have something that looks like a shadow. e.g. the contained one, ins't contained anymore.

src/components/public/diamondgs.png
akademy, line 1: This file is just grey for me. 
(In future I'd prefer a SVG here, and I know Diamond has a good one already.)

src/themes/DiamondTheme.ts
```diff
@@ -1,47 +1,86 @@
...
+            color: theme.palette.primary.contrastText,
+            backgroundColor: theme.palette.primary.dark,
+            "&:hover": {
+              backgroundImage: "linear-gradient(rgb(0 0 0/30%) 0 0)",
+              "&:disabled": {
+                backgroundColor: theme.palette.primary.light,
+              },
+            },
+          },
+          "&.onBlue": {
```
akademy, line 61: I wonder if this was supposed to say "onBlur" !

src/components/public/diamondgs.png
VictoriaBeilsten-Edmands: Yes, I used the same (plain grey) file from web-ui-components. I've only seen a Diamond jpg so this can be replaced with svg.

src/themes/DiamondTheme.ts
VictoriaBeilsten-Edmands: I think it's for when the background is (Diamond) blue.

src/components/Navbar.stories.tsx
VictoriaBeilsten-Edmands: Yes, I agree.

src/themes/DiamondTheme.ts
VictoriaBeilsten-Edmands: I don't think we really want this border anyway. Could remove this line and then add `"&.MuiButton-contained": {},` below:

`root: ({ theme }) => ({
          textTransform: "none",
          "&.MuiButton-contained": {},
          "&.default": {`
          
Something like this would prevent the "contained" variant being overridden and can then be modified as we want. It still applies the text transform (text not changed to uppercase) to everything.

src/themes/DiamondTheme.ts
akademy: I talked with Guilherme and he think it's a typo in his original code.

VictoriaBeilsten-Edmands: I didn't see it used anywhere with onBlur, I can change it to something more descriptive for how it's being used here.

PR #3, opened 2024-11-20
Add VisitInput component

VictoriaBeilsten-Edmands:


src/components/VisitInput.stories.test.tsx
akademy, line 1: Rename without "stories."!

src/components/VisitInput.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const [visitText, setVisitText] = useState(
+    visit
+      ? `${visit.proposalCode}${visit.proposalNumber.toFixed(
+          0
+        )}-${visit.number.toFixed(0)}`
+      : ""
+  );
+  const [isValid, setIsValid] = useState(true);
+
+  const visitRegex = /^([a-z]{2})([1-9]\d*)-([1-9]\d*)/;
```
akademy, line 30: Can we have a test for this regex. Something that just tests a group of valid and invalid ones.

akademy (commented):
This is a cool idea.

Some initial thoughts:

- Would be nice to get the input working on it's own without the submit perhaps with `<input type="text" pattern="/^([a-z]{2})([1-9]\d*)-([1-9]\d*)/">` (no idea how easy that is in react/mui) , could then be used in more complicated forms.
- Maybe this component should be called VisitForm? or VisitSubmit? VisitInputSubmit?
- Should we move the regex and code to a different file? It maybe useful for other components to manipulate the visit string.

VictoriaBeilsten-Edmands (commented):
How about I put the visit regex in a separate file like utils/diamond.ts. Then separate it into a VisitInput component which would just be the text input using that visit regex. Then a second component which includes the submit buttton. That should keep everything as reusable as possible.

akademy:
> How about I put the visit regex in a separate file like utils/diamond.ts. Then separate it into a VisitInput component which would just be the text input using that visit regex. Then a second component which includes the submit buttton. That should keep everything as reusable as possible.

Yeah, perfect.

akademy:
Sorry, closed by mistake!

src/components/VisitInput.stories.test.tsx
VictoriaBeilsten-Edmands: Renamed

src/components/VisitInput.tsx
VictoriaBeilsten-Edmands: I've added tests and also modified the regex slightly to prvent any extra trailing groups.

VictoriaBeilsten-Edmands (commented):
I've moved the visit regex and functions for parsing visit strings etc into utils/diamond.ts.
I've separated the visit input text box from the submit button, but ultimately left it in the same file without separate stories as VisitInput returns just the TextField if no onSubmit prop is given. This can be moved around at some point if we want.

akademy:
Looks good to me.

PR #5, opened 2024-11-21
Add a Storybook theme mode switcher (dark or light)

akademy:
Added a button to switch between the two modes, dark and light.

Mostly it's just boiler plate. The `ThemeSwapper` component wraps Storybook and updates MUI's theme setting when the button is pressed. 

It defaults to light.


VictoriaBeilsten-Edmands (commented):
I like the theme switching. Could we get a theme switching button on the top of the page to switch the theme of all components at once? There is a storybook theme switching button in the header, but that is for switching the theme of the storybook, not the components. I'm not sure we necessarily want to tie this to that button.

akademy:
> I like the theme switching. Could we get a theme switching button on the top of the page to switch the theme of all components at once? There is a storybook theme switching button in the header, but that is for switching the theme of the storybook, not the components. I'm not sure we necessarily want to tie this to that button.

I looked into this, and it was surprisingly easy to add additional buttons! I also added a theme switcher as well as the theme mode switcher.

PR #9, opened 2024-11-28
Use peer dependencies

gfrn:
- Moved React, Emotion, MUI, and other peer dependencies to `peerDependencies`. This will reduce the likelihood of clashes when setting up hooks in applications that have `sci-react-ui` as a dependency. Besides, it's common practice, and since NPM 7, these are installed automatically.
- Moved dev dependencies like `typescript` and test dependencies to `devDependencies` to reduce the production bundle size
- Removed `@material-ui/core`. This library has been deprecated since 2021 and has been superseded by `@mui/material`, so it should not be in the dependencies, especially since it's not being used

We should also consider substituting Jest for Vitest, as Jest is much more likely to clobber the DOM, performs worse, and has weird behaviour that is not compatible with other test libraries like MSW. Besides, Jest uses a number of deprecated libraries.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,24 +20,27 @@
   "module": "dist/index.esm.js",
   "types": "dist/index.d.ts",
   "dependencies": {
-    "@emotion/react": "^11.13.3",
-    "@emotion/styled": "^11.13.0",
     "@mui/icons-material": "^6.1.7",
-    "@mui/material": "^6.1.7",
-    "@testing-library/jest-dom": "^6.6.3",
+    "react-icons": "^5.3.0"
+  },
+  "peerDependencies": {
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
-    "react-icons": "^5.3.0",
-    "typescript": "^5.6.3",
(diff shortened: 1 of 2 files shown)


akademy:
Change looks good here.

gfrn:
I mentioned this in one of the PR comments for the policy, but I don't think it ever got merged in

That being said, I'm happy to make a PR swapping out Jest for Vitest if that's alright

akademy:
I'd be fine with that. 

If you wanted to write some tests too, we are missing them for the NavBar (which you mostly wrote, anyway :wink:  )...!

gfrn:
Alright! I'll wait on some feedback from Victoria before moving on with the move to Vitest, but I'll start copying over the navbar tests as soon as I get a bit more time

VictoriaBeilsten-Edmands:
I'm also happy with a switch to Vitest. If you want to copy any tests over that would be great too. Thanks!

gfrn:
Great! Could someone merge this to main, if all is OK?

PR #10, opened 2024-12-06
Standardise exports

akademy:
Hello.

Just a tweak and standardising the exports. We had a problem importing the VisitInput into one of our projects.

Pretty basic changes.

### DIFF
--- a/src/components/VisitInput.tsx
+++ b/src/components/VisitInput.tsx
@@ -89,6 +89,5 @@ const VisitInput: React.FC<VisitInputProps> = ({
   );
 };
 
-export default VisitInput;
-export { VisitInputText };
+export { VisitInput, VisitInputText };
 export type { Visit, VisitInputTextProps, VisitInputProps };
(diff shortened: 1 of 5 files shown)


src/components/VisitInput.tsx
VictoriaBeilsten-Edmands, line 1: Need to update VisitInput import in `VisitInput.stories.tsx` to `import { VisitInput } from "./VisitInput";`

src/components/VisitInput.tsx
```diff
@@ -89,6 +89,5 @@ const VisitInput: React.FC<VisitInputProps> = ({
   );
 };
 
-export default VisitInput;
```
VictoriaBeilsten-Edmands, line 92: Need to update VisitInput import in VisitInput.stories.tsx to import { VisitInput } from "./VisitInput";

akademy: Good spot! Have updated. Ta.

VictoriaBeilsten-Edmands: Oh, it's in `VisitInput.test.tsx` too.

PR #18, opened 2025-01-06
Colour scheme switcher.

akademy:
Just a button to switch between dark and light color schemes. 

(We've done all that hard work, so we HAVE to make it usable!)


VictoriaBeilsten-Edmands (approved):
Great!

PR #25, opened 2025-01-10
Add navbar tests, fix Jest/Babel configs

gfrn:
* Add navbar tests
* Jest was declared as a module, but had no default export, this makes the Jest config a module. Babel has been made CommonJS, as it is the standard with this version of Babel

Next, I think I'll add workflow actions for automatically running unit tests, and possible move to Vitest!


VictoriaBeilsten-Edmands (commented):
Thanks for adding these tests and for the config updates.
As we've got rid of `"type": "module"` in `package.json` do we still need to change `babel.config.js` to `babel.config.cjs`?

gfrn:
I based this off an older master branch, but I'll test it without `type: module` later tomorrow

gfrn:
That should do it, but it's not wholly necessary - it still uses CJS module exports than proper `export`

VictoriaBeilsten-Edmands (approved):
Thanks.

PR #38, opened 2025-03-21
Breadcrumb to accept Array<object> 

RhysMAdey:
User will be able to pass and object array into 'path'. They will be able to determine the name the users see and the link when they press that breadcrumb item. e.g.

path=[ {name: 'link 1', href:'www.link1.com'}, {name: 'link 2', href: 'www.link2.com' }]

Breaedcrumb comp 100% tested:
Screenshot From 2025-03-21 15-03-51

### DIFF
--- a/src/components/Breadcrumbs.test.tsx
+++ b/src/components/Breadcrumbs.test.tsx
@@ -2,16 +2,12 @@ import { render, RenderResult } from "@testing-library/react";
 import { Breadcrumbs, getCrumbs } from "./Breadcrumbs";
 import "@testing-library/jest-dom";
 
+const defaultArrayObject = [
+  { name: "first", href: "first/is/this" },
+  { name: "second", href: "second" },
+  { name: "last one", href: "last one" },
+];
 describe("Breadcrumbs", () => {
-  const crumbFirst = "first",
-    crumbFirstTitle = "First",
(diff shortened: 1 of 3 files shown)


src/components/Breadcrumbs.test.tsx
```diff
@@ -2,16 +2,20 @@ import { render, RenderResult } from "@testing-library/react";
...
 
+const crumbFirst = "first",
+  crumbFirstTitle = "First",
+  crumbSecond = "second",
+  crumbSecondTitle = "Second",
+  crumbLast = "last one",
+  crumbLastTitle = "Last one",
+  defaultStringPath = `/${crumbFirst}/${crumbSecond}/${crumbLast}`,
+  defaultArrayPath = [crumbFirst, crumbSecond, crumbLast],
+  defaultArrayObject = [
```
akademy, line 13:
```suggestion
  defaultArrayObject: CustomLink[]  = [
```

akademy (commented):
This looks good. One tiny improvement.

src/components/Breadcrumbs.test.tsx
RhysMAdey: Pushed

akademy (approved):

PR #39, opened 2025-04-04
Fix movement on hover when only links.

akademy:
The menu items would move up and down when there's only the links set. This is a fix for that.

Also, some changes to add to centre the menu (so it's more like a traditional footer!). 

And a little refactor of the storybook file.


RhysMAdey:
Would we want the links centred like the copyright info?

Screenshot From 2025-04-07 15-13-57

akademy:
> Would we want the links centred like the copyright info?
> 
> Screenshot From 2025-04-07 15-13-57

Should be fixed in latest version.

RhysMAdey (approved):

PR #40, opened 2025-04-08
Use named imports for MUI components

gfrn:
In some cases, using default exports can cause builds/tests to fail:

`Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.`

 This PR makes all MUI component imports named imports, which fixes these issues.


akademy (approved):
Thanks. Looks good. I'll merge it in.

Issue #35, opened 2025-01-30
NavBar has some strange positioning of logo and menu on different sizes.


akademy:
The logo on the left always has a large left space next to it, it should be much closer to the edge (though perhaps not on very wide screens)
![Image](https://github.com/user-attachments/assets/725b5016-9207-4b57-92a9-5ab5cd32a381)

When the Links become a menu on small screens, the menu also has a large gap to it's left.
![Image](https://github.com/user-attachments/assets/a909fe59-ae6b-48f4-80e2-55602ea2c411)

Also, when there are additional items on the NavBar (like the User), it doesn't show them on small screens (this may partly be a problem with the User control not becoming smaller when needed)
![Image](https://github.com/user-attachments/assets/2e046ca9-54f2-45f8-845f-53ac3a04850a)

akademy:
Fixed.

Issue #33, opened 2025-01-27
Not clear what version of SciReactUI Storybook is running.


akademy:
It should display the version it is built against.

akademy:
Storybook now displays the version in the package.json it is built against.

PR #41, opened 2025-04-08
Fix layout of navbar at various different widths.

akademy:
Hi.

The positioning of the logo and buttons in the NavBap has always looked a bit off. This corrects the positioning on multiple screen widths. 

This is to correct issue #35

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,10 +4,14 @@ SciReactUI Changelog
 [vNext] - date
 --------------------
 
+### Added
+- Breadcrumbs takes object for total control over names and links 
+
 ### Fixed
-- Stopped flicker between themes when starting an app in dark mode.
+- Stopped flicker between colour modes when starting an app in dark mode.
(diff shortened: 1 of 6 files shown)


RhysMAdey:
Very minor comment and if it is a lot of effort to resolve then I would happily let this comment be ignored:

Do we want the outer left and outer right spacing to be the same when just links and user?
Screenshot From 2025-04-09 10-28-33

Something like:
Screenshot From 2025-04-09 10-31-20

Again, very minor so if you are happy with what you have then thats ok

RhysMAdey:
Other than above comment - all looks good and works as expected for me

akademy:
Hi @RhysMAdey ,

Fair point above. I've made a fix for that.

akademy:
Also partially improves https://github.com/DiamondLightSource/sci-react-ui/issues/33

RhysMAdey (approved):

PR #59, opened 2025-05-02
Add react router links to Breadcrumbs

### LINKED ISSUE #56: Support routing library navigation in Breadcrumbs component
As a developer writing a single page app, I would like to use a routing library (e.g. React Router) when navigating using the `Breadcrumbs` component. This would prevent page refreshes/app reloads when clicking on path elements in the `Breadcrumbs` component.

Perhaps we could have something like
```tsx
import {Link as RouterLink} from "my-routing-library";
...
    <Breadcrumbs path="/some/example/path" linkProps={{component:RouterLink}}/>
```

VictoriaBeilsten-Edmands:
Adds LinkComponent prop for page routing. Closes #56.
Reorganises components to mirror Storybook component structure.
Fixes type issue with image in Footer and Navbar.
I had to add yet another config file to get the tests running due to an issue with react-router-dom v7. See [here](https://stackoverflow.com/questions/79332161/referenceerror-textencoder-is-not-defined-after-react-router-dom-upgrade-to-7) and [here](https://remarkablemark.org/blog/2025/02/02/fix-jest-errors-in-react-router-7-upgrade/#:~:text=This%20error%20occurs%20when%20you%20bump%20react-router-dom%20from,install%20text-encoding%3A%20Update%20your%20Jest%20config%20with%20setupFiles%3A).


package.json
```diff
@@ -24,13 +24,15 @@
     "storybook:publish": "gh-pages -b storybook/publish -d storybook-static"
   },
   "dependencies": {
-    "react-icons": "^5.3.0"
+    "react-icons": "^5.3.0",
+    "react-router-dom": "^7.5.3",
```
akademy, line 28: This should be under peerDependencies.

package.json
```diff
@@ -24,13 +24,15 @@
...
+    "storybook-addon-remix-react-router": "3"
```
akademy, line 29: And this under devDependencies

src/components/Breadcrumbs.tsx
```diff
@@ -3,16 +3,15 @@ import {
   Breadcrumbs as Mui_Breadcrumbs,
   BreadcrumbsProps as Mui_BreadcrumbsProps,
   Container,
-  Link,
+  Link as MuiLink,
```
akademy, line 6: Can we make this consistent with an underline, as e.g. Mui_BreadcrumbsProps please

src/components/Breadcrumbs.tsx
```diff
@@ -79,22 +78,29 @@ const Breadcrumbs = ({
           sx={{ color: theme.palette.primary.contrastText }}
           {...muiBreadcrumbsProps}
         >
-          <Link key={"crumb-0"} underline="hover" color="inherit" href="/">
+          <MuiLink
+            key={"crumb-0"}
+            underline="hover"
+            color="inherit"
+            component={Link}
```
akademy, line 85: I did wonder whether this should be NavLink rather than Link but I think this is right for breadcrumbs.

akademy (changes requested):
This didn't work "out of the box" for me when I used it in my external app, because I didn't have `<BrowserRouter>` added.

I'm not sure how to handle this. Maybe we could add it to some-kind of `<SciReactProvider/>` (which could also embed the theme) so that just the one line can be added in external apps. Something like:

```React
const SciReactUIProvider = function ({children}) {
  return (
    <ThemeProvider>
      <BrowserRouter>
        {children}
      </BrowserRouter>
    </ThemeProvider>
  );
};
```

The alternative is to just add it to the docs but I like to hide, behind the scenes, as much as we can  What do you think?

VictoriaBeilsten-Edmands:
We could create the SciReactUIProvider which would leave the Breadcrumbs component unchanged and would leave the option for people to use e.g. RouterProvider in place of BrowserProvider together withe the ThemeProvider. I should still update the docs with something.

akademy:
I thought we still had to change Breadcrumbs to use ReactRouter's `Link` though. Or does it just pick up MUI's version?

VictoriaBeilsten-Edmands:
My comment was unclear! I mean leave the Breadcrumbs component unchanged from how it is in this PR.

src/components/Breadcrumbs.tsx
VictoriaBeilsten-Edmands: Agreed.

changelog.md
```diff
@@ -6,13 +6,13 @@ SciReactUI Changelog
 --------------------
 
 ### Added
-- 
+- SciReactUIProvider to use ThemeProvider and BrowserRouter.
```
akademy, line 9: Maybe highlight this as a breaking change, as it won't work without the <BrowserRouter>. And bump the version to 0.2.0 (also in packages.json)

src/themes/SciReactUIProvider.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
+import { BrowserRouter } from "react-router-dom";
+import { ThemeProvider, ThemeProviderProps } from "./ThemeProvider";
+
+type SciReactUIProviderProps = React.PropsWithChildren<ThemeProviderProps>;
```
akademy, line 5: Should we also add `BrowserRouterProps` to the type so we can add them with `<BrowserRouter{...props.router}>` or similar?

akademy (changes requested):
Looks good. Just a few minor suggestions.

akademy:
The `SciReactUIProvider` doesn't really fit in the "Themes" folder... but I don't really know the best place to put it...

VictoriaBeilsten-Edmands:
> The `SciReactUIProvider` doesn't really fit in the "Themes" folder... but I don't really know the best place to put it...

I was a bit stuck on where to put it.

akademy:
Maybe it's time to reorganise the component folder... controls, sections, themes, general ?

akademy:
> > The `SciReactUIProvider` doesn't really fit in the "Themes" folder... but I don't really know the best place to put it...
> 
> I was a bit stuck on where to put it.

Maybe it's time to reorganise the component folder... controls, sections, themes, general ?

gfrn:
> > Are we sure we want to add a dependency to a specific routing library? Or has Diamond made a decision on which we should use?
> 
> As far as I know there's no Diamond-wide decision on which we should use. I do think react-router is a good choice and what we should use here. I don't see a reason to diverge with multiple options.

I think allowing people to do component "impersonation" in the breadcrumbs component is not going to lead to diversion, it would be useful for the following reasons:

- If someone decides to use a non-`BrowserRouter` provider for `react-router`, then they are able to do so 
- We are not locked to one specific version of `react-router`, and being able to decouple updating `react-router` with updating `sci-react-ui` is really useful, because it might not be trivial for all people (especially if they are using/want to use version specific features, and going from v6 to v7 was not easy if you're using advanced features)
- Not everyone is going to use `react-router` anyway, some people might elect to use native browser navigation, or in my case, Next's own native navigation
- If we choose to move to a different routing library in the future (like Tanstack Router), the switch is much easier, and we're not tying our UI to a routing library
- This decreases the number of peer/direct dependencies, so when security updates come out, people won't have to wait on us updating `sci-react-ui`, so this is a safer option

VictoriaBeilsten-Edmands:
> > > Are we sure we want to add a dependency to a specific routing library? Or has Diamond made a decision on which we should use?
> > 
> > 
> > As far as I know there's no Diamond-wide decision on which we should use. I do think react-router is a good choice and what we should use here. I don't see a reason to diverge with multiple options.
> 
> I think allowing people to do component "impersonation" in the breadcrumbs component is not going to lead to diversion, it would be useful for the following reasons:
> 
>     * If someone decides to use a non-`BrowserRouter` provider for `react-router`, then they are able to do so
> 
>     * We are not locked to one specific version of `react-router`, and being able to decouple updating `react-router` with updating `sci-react-ui` is really useful, because it might not be trivial for all people (especially if they are using/want to use version specific features, and going from v6 to v7 was not easy if you're using advanced features)
> 
>     * Not everyone is going to use `react-router` anyway, some people might elect to use simpler native browser navigation, or in my case, Next's own native navigation
> 
>     * If we choose to move to a different routing library in the future (like Tanstack Router), the switch is much easier, and we're not tying our UI to a routing library
> 
>     * This decreases the number of peer/direct dependencies, so when security updates come out, people won't have to wait on us updating `sci-react-ui`, so this is a safer option
> 
>     * It's not a "free" win, but it's at least technically feasible, and MUI already does this for plenty of components (in fact, this is a feature which is being used in this PR)

I could change the Breadcrumbs component and pass a LinkComponent e.g.
```
interface BreadcrumbsProps {
  path: string | string[] | CustomLink[];
  linkComponent: React.ElementType
  rootProps?: PaperProps;
  muiBreadcrumbsProps?: Mui_BreadcrumbsProps;
}
```
Then this can be used with whatever routing library e.g. Link from react-router-dom/ next etc.
We wouldn't have the SciReactUIProvider if we want to keep the dependency out of the library, so that would be removed.

VictoriaBeilsten-Edmands:
> > > The `SciReactUIProvider` doesn't really fit in the "Themes" folder... but I don't really know the best place to put it...
> > 
> > 
> > I was a bit stuck on where to put it.
> 
> Maybe it's time to reorganise the component folder... controls, sections, general ?

Shall we go with component/controls and component/navigation to mirror the storybook. I quite like those categories.

gfrn:
> > > > Are we sure we want to add a dependency to a specific routing library? Or has Diamond made a decision on which we should use?
> > > 
> > > 
> > > As far as I know there's no Diamond-wide decision on which we should use. I do think react-router is a good choice and what we should use here. I don't see a reason to diverge with multiple options.
> > 
> > 
> > I think allowing people to do component "impersonation" in the breadcrumbs component is not going to lead to diversion, it would be useful for the following reasons:
> > ```
> > * If someone decides to use a non-`BrowserRouter` provider for `react-router`, then they are able to do so
> > 
> > * We are not locked to one specific version of `react-router`, and being able to decouple updating `react-router` with updating `sci-react-ui` is really useful, because it might not be trivial for all people (especially if they are using/want to use version specific features, and going from v6 to v7 was not easy if you're using advanced features)
> > 
> > * Not everyone is going to use `react-router` anyway, some people might elect to use simpler native browser navigation, or in my case, Next's own native navigation
> > 
> > * If we choose to move to a different routing library in the future (like Tanstack Router), the switch is much easier, and we're not tying our UI to a routing library
> > 
> > * This decreases the number of peer/direct dependencies, so when security updates come out, people won't have to wait on us updating `sci-react-ui`, so this is a safer option
> > 
> > * It's not a "free" win, but it's at least technically feasible, and MUI already does this for plenty of components (in fact, this is a feature which is being used in this PR)
> > ```
> 
> I could change the Breadcrumbs component and pass a LinkComponent e.g.
> 
> ```
> interface BreadcrumbsProps {
>   path: string | string[] | CustomLink[];
>   linkComponent: React.ElementType
>   rootProps?: PaperProps;
>   muiBreadcrumbsProps?: Mui_BreadcrumbsProps;
> }
> ```
> 
> Then this can be used with whatever routing library e.g. Link from react-router-dom/ next etc. We wouldn't have the SciReactUIProvider if we want to keep the dependency out of the library, so that would be removed.

I think this is a nice solution (to allow users to set both the link component and the prop each breadcrumb would use in that link component), and removing the provider surely would make things easier in the future!

akademy (approved):

PR #61, opened 2025-05-21
Add routing library support to FooterLink, NavLink & Navbar

VictoriaBeilsten-Edmands:
They all should keep the same static link behaviour as before if not given `linkComponent` and `to` and instead given href.


src/components/navigation/Footer.tsx
```diff
@@ -42,15 +42,36 @@ const FooterLinks = ({ children, ...props }: FooterLinksProps) => {
...
+  linkComponent,
+  to,
+  href,
+  ...props
+}: FooterLinkProps) => {
   const theme = useTheme();
 
+  const shouldUseLinkComponent = linkComponent && to;
+
+  const linkProps = shouldUseLinkComponent
```
akademy, line 63: Do we need this? Can't we just pass through what's been set on FooterLink? i.e. Get the user to decide what they want to use (and if they need "href" or "to").

akademy: I think we just need to and linkComponent to the LinkProps (and href optional, if it isn't already)

src/components/navigation/Navbar.tsx
```diff
@@ -25,17 +24,39 @@ interface NavLinksProps {
...
+  linkComponent,
+  to,
+  href,
+  ...props
+}: NavLinkProps) => {
   const theme = useTheme();
 
+  const shouldUseLinkComponent = linkComponent && to;
+
+  const linkProps = shouldUseLinkComponent
```
akademy, line 48: Don't think we need this one either. (But we do still need it on linkComponent on NavbarProps)

src/components/navigation/Navbar.tsx
```diff
@@ -126,7 +147,7 @@ const BoxStyled = styled(Box)<BoxProps>(({ theme }) => ({
 /**
  * Basic navigation bar. Can be used with `NavLinks` and `NavLink` to display a responsive list of links.
  */
-const Navbar = ({ children, logo, ...props }: NavbarProps) => {
+const Navbar = ({ children, logo, linkComponent, ...props }: NavbarProps) => {
   const theme = useTheme();
   let resolvedLogo: ImageColorSchemeSwitchType | null | undefined = null;
```
akademy, line 152: A side note... This won't ever be undefined as we are setting it to null.

akademy (changes requested):
A few things...

src/components/navigation/Footer.tsx
VictoriaBeilsten-Edmands: I could do it that way. I'd need to rename linkComponent to component to pass directly to Link. It would mean there's no fallback to href if the wrong props are used e.g. if component and href are passed together without to.

src/components/navigation/Navbar.tsx
VictoriaBeilsten-Edmands: normal in theme.logos?.normal is typed to `'ImageColorSchemeSwitchType | undefined'`, so it can end up `undefined` with
`resolvedLogo = theme.logos?.normal;`

src/components/navigation/Footer.tsx
akademy: It deleted my message...

src/components/navigation/Navbar.tsx
akademy: Leave it.

src/components/navigation/Navbar.tsx
akademy: I see.

(Feels like this package is getting more complicated every day... )

RhysMAdey (approved):

akademy (approved):

PR #68, opened 2025-06-03
Add Navbar slots

VictoriaBeilsten-Edmands:
Adds slots to Navbar: `logoLeftSlot`, `logoRightSlot`, `centreSlot`, `rightSlot`.
If a logo is defined (either via the `logo` prop or from the theme), the layout will arrange elements in the following order from left to right: `logoLeftSlot`, `logo`, `logoRightSlot`, `rightSlot`.
The `centreSlot` is absolutely positioned at 50% horizontally, which means it stays centered regardless of the content on the left or right. However, if the content in the left or right slots is too wide, it may overlap with the centre slot.

Existing uses of Navbar will need to update anything on the far right to be in `rightSlot`. Children should behave as before. By changing slot props, there should be a lot more freedom over the layout.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -105,9 +105,13 @@ function App() {
 }
 export default App;
 ```
+The Navbar component supports multiple slot props for flexible layout customization: leftSlot, centreSlot, rightSlot.
+If a logo is defined (either via the logo prop or from the theme), the layout will arrange elements in the following order from left to right: logo, leftSlot, rightSlot.
+The centreSlot is absolutely positioned at 50% horizontally, which means it stays centered regardless of the content on the left or right. However, if the content in the left or right slots is too wide, it may overlap with the centre slot.
 
-There are various other components, here's an example of how to use the NavBar:
+Any children passed to the Navbar (NavLinks in the following example) will be placed in a horizontal Stack after the leftSlot.
 
+To control the width of the Navbar contents, containerWidth can be passed which sets the maxWidth property of a Container. This can be set to false to not constrain the contents or to a Breakpoint e.g. 'sm' where 'lg' is the default.
 ```js
 import { Container, Typography } from "@mui/material";
 import { Navbar, NavLink, NavLinks } from "@diamondlightsource/sci-react-ui";
@@ -115,17 +119,18 @@ import { Navbar, NavLink, NavLinks } from "@diamondlightsource/sci-react-ui";
 function App() {
   return (
     <>
-      <Navbar>
+      <Navbar
+        leftSlot={<Typography>left</Typography>}
+        centreSlot={<Typography>centre</Typography>}
+        rightSlot={<Typography>right</Typography>}
+        containerWidth={false}
+      >
         <NavLinks key="links">
           <NavLink href="#" key="first">
             A link
           </NavLink>
         </NavLinks>
       </Navbar>
-      <Container>
-        <Typography variant="h2">Scientific UI Collection</Typography>
-        <Typography>A collection of science based React components.</Typography>
-      </Container>
     </>
   );
 }
--- a/src/components/navigation/Navbar.stories.tsx
+++ b/src/components/navigation/Navbar.stories.tsx
@@ -19,43 +19,43 @@ type Story = StoryObj<typeof meta>;
 
 export const All: Story = {
   args: {
-    children: [
-      <NavLinks key="links">
-        <NavLink href="#Mercury" key="mercury">
-          Mercury
-        </NavLink>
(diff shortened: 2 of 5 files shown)


src/components/navigation/Navbar.stories.tsx
```diff
@@ -161,6 +176,54 @@ export const CustomChildElement: Story = {
   },
 };
 
+export const LinksInSlot: Story = {
+  args: {
+    rightSlot: [
```
akademy, line 181: Can we get these using just `React.ReactElement<LinkProps>` rather than the list `React.ReactElement<LinkProps>[]` 

And  use the style like below, I think it's "more React-like":
```
<>
      <User
        key="user"
        onLogin={() => {}}
        onLogout={() => {}}
        user={{ name: "Name", fedid: "FedID" }}
        color={"white"}
      />
      <ColourSchemeButton key="colourScheme" />
</>
```

src/components/navigation/Navbar.stories.tsx
```diff
@@ -161,6 +176,54 @@ export const CustomChildElement: Story = {
...
+        <NavLink href="#Mars" key="mars">
+          Mars
+        </NavLink>
+      </NavLinks>,
+    ],
+    logo: "theme",
+  },
+};
+
+export const AllSlots: Story = {
```
akademy, line 211: In Storybook and clicking on "Show Code" this wierdly generates:

```
<Navbar
  centreSlot={<React.ForwardRef sx={{color: 'white'}}>centre</React.ForwardRef>}
  logo="theme"
  logoLeftSlot={<React.ForwardRef sx={{color: 'white'}}>logo left</React.ForwardRef>}
  logoRightSlot={<React.ForwardRef sx={{color: 'white'}}>logo right</React.ForwardRef>}
  rightSlot={<React.ForwardRef sx={{color: 'white'}}>right</React.ForwardRef>}
>
  <React.ForwardRef
    sx={{
      color: 'white'
    }}
  >
    children
  </React.ForwardRef>
</Navbar>
```

Any ideas why it is using `React.ForwardRef`?

src/components/navigation/Navbar.tsx
```diff
@@ -185,9 +208,21 @@ const Navbar = ({ children, logo, linkComponent, ...props }: NavbarProps) => {
               </Box>
             </Link>
           )}
+          {logoRightSlot}
           {children}
```
akademy, line 212: Maybe it makes more sense to add the children by default into the logoRightSlot, rather than a separate "special" slot. 

(Then maybe it should be called "leftSlot" to mirror "rightSlot...)

readme.md
```diff
@@ -115,17 +118,18 @@ import { Navbar, NavLink, NavLinks } from "@diamondlightsource/sci-react-ui";
 function App() {
   return (
     <>
-      <Navbar>
+      <Navbar
+        logoLeftSlot={<Typography>logo left</Typography>}
+        logoRightSlot={<Typography>logo left</Typography>}
```
akademy, line 123:
```suggestion
        logoRightSlot={<Typography>logo right</Typography>}
```

src/components/navigation/Navbar.stories.tsx
akademy: This is a comment on some of the other examples too.

akademy (changes requested):

src/components/navigation/Navbar.stories.tsx
VictoriaBeilsten-Edmands: Updated these.

src/components/navigation/Navbar.stories.tsx
VictoriaBeilsten-Edmands: I'm not seeing that. Do you still see it with the latest version?

src/components/navigation/Navbar.tsx
VictoriaBeilsten-Edmands: Removed logoLeftSlot and now leftSlot content gets children too.

src/components/navigation/Navbar.stories.tsx
```diff
@@ -99,23 +101,28 @@ export const RouterLinks: Story = {
...
+      <>
+        <NavLinks key="links">
+          <NavLink href="#" key="first">
+            First
+          </NavLink>
+          <NavLink href="#" key="second">
+            Second
+          </NavLink>
+        </NavLinks>
+        ,
```
akademy, line 123: Some commas hanging around.

akademy (approved):

PR #79, opened 2025-06-27
Vertically align elements of Breadcrumbs

VictoriaBeilsten-Edmands:
This vertically aligns the text and home icon with the chevrons.


douglaswinter (approved):
Looks much better, thank you. There is also an argument to be made for moving the selected item too (since it isn't just bold, but bigger), but that is more subjective.

PR #81, opened 2025-06-27
Rename color to colour

### LINKED ISSUE #78: ImageColorSchemeSwitch needs to be renamed to ImageColourSchemeSwitch
We agreed to use British English for components. We'll need to switch this component over.

VictoriaBeilsten-Edmands:
ImageColorSchemeSwitch, ImageColorSchemeSwitchType and ImageColorSchemeSwitchProps renamed to ImageColourSchemeSwitch, ImageColourSchemeSwitchType and ImageColourSchemeSwitchProps. User component color prop renamed to colour.
Renamed color to colour in stories.

Closes #78

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,18 @@ SciReactUI Changelog
 ====================
 
 
+[v0.2.1alpha] - 2025-??-??
+--------------------
+
+### Added
+-
+
+### Fixed
+- 
+
+### Changed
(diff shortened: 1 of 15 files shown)


jacob720 (approved):
This looks good. I was wondering about something similar recently, as my `ScrollableImages` component takes a prop called `backgroundColor`. Initially, I named it `backgroundColour`, but I changed it so it was consistent with the `Box` prop naming that the prop was ultimately being passed to. What's the best convention to use in this situation?

VictoriaBeilsten-Edmands:
> This looks good. I was wondering about something similar recently, as my `ScrollableImages` component takes a prop called `backgroundColor`. Initially, I named it `backgroundColour`, but I changed it so it was consistent with the `Box` prop naming that the prop was ultimately being passed to. What's the best convention to use in this situation?

If it's being passed directly to something else e.g. passing sx props to Box for example, probably best to stick with color.

PR #92, opened 2025-07-24
Only include relevant files in npm package

### LINKED ISSUE #75: Several unnecessary files are being output into the NPM package.
Several config files are being unnecessarily added to the npm package:
- babel.config.js
- eslint.config.js
- jest.setup.js
- tsconfig.eslint.json

Also some generated test and storybook files:
- /.../{component-name}.test.d.ts
- /.../{component-name}.stories.d.ts

 There may be other files we can ignore too.

jacob720:
Fixes #75 
Added a files section to `package.json` so that only the relevant files are included in the npm package. 
Currently I've included `/src/` but I'm not sure this is necessary? I could remove this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,11 @@
 {
   "name": "@diamondlightsource/sci-react-ui",
   "version": "0.2.1-alpha",
+  "files": [
+    "dist/",
+    "src/",
+    "index.js"
+  ],
   "description": "A theme and component library to make websites at scientific institutions simple to create.",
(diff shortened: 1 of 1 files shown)


akademy:
> Fixes #75 Added a files section to `package.json` so that only the relevant files are included in the npm package. Currently I've included `/src/` but I'm not sure this is necessary? I could remove this.

Could you check whether src/ is necessary? I'm not sure either.

PR #96, opened 2025-08-12
Extend ScrollableImages to view tiffs

jacob720:
I've extended the ScrollableImages component so that it can handle multi-image tiffs. It will split the frames into seperate png images which can then be displayed by the browser.

### DIFF
--- a/src/components/controls/ScrollableImages.tsx
+++ b/src/components/controls/ScrollableImages.tsx
@@ -34,7 +34,7 @@ const ScrollableImages = ({
   const [extractedImages, setExtractedImages] = useState<ImageInfo[]>([]);
 
   useEffect(() => {
-    async function processImages() {
+    (async () => {
       const inputImages = Array.isArray(images) ? images : [images];
       let result: ImageInfo[] = [];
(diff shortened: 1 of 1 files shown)


VictoriaBeilsten-Edmands:
There's a warning
```
Warning: Styled(div) contains an input of type number with both value and defaultValue props. Input elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both). Decide between using a controlled or uncontrolled input element and remove one of these props.
```

It's from having defaultValue and onChange here
```
        {renderNumbers && (
          <Box sx={{ display: "flex" }}>
            <Box
              data-testid="numeration"
              component="input"
              type="number"
              value={numberValue}
              defaultValue={""}
              onChange={handleNumberChange}
              onKeyDown={handleNumberEnter}
```

If it's controlled input you have value and onChange and if it's uncontrolled you have defaultValue but you can only have one or the other.

https://react.dev/reference/react-dom/components/input

jacob720:
Thanks, I've removed the defaultValue prop, as it should be a controlled input.

PR #98, opened 2025-08-28
Replacing JsonForms readonly inputs with accessible text.

akademy:
This replaces the default readonly text inputs in JsonForms. There are two base controls, DataBox and DataCell are reused in all renderers and cells.

I've added a description of how to use them too StoryBook, but the default StoryBook layout was confusing, so I've changed the ordering, and renamed the group "SciReactUI" to "Components" which affected several files in a small way.

I also tweaked the StoryBook Introduction and updated the Changelog file.

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -75,6 +75,18 @@ const preview: Preview = {
     },
     backgrounds: { disable: true },
     layout: "fullscreen",
+    options: {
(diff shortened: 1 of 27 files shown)


RhysMAdey:
Added in to own app and looks as expected for me

Screenshot From 2025-08-28 16-33-33

PR #112, opened 2025-10-16
Switch from jest to vitest

akademy:
Switching from Jest to Vitest.

Some CommonJs/ESM/Blah blah javascript was causing a minefield to test with Jest. It was easier to switch to this better system than figure out the black magic...

This is JUST the code to switch to ViTest. Everything should be running the same, but perhaps a bit quicker.

(Also, @gfrn was always asking for this!)

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -20,6 +20,7 @@ export default [
     "**/*.md",
     "*.config.js",
     "rollup.config.mjs",
+    "vitest.setup.ts"
   ]},
   js.configs.recommended,
   ...compat.extends("plugin:@typescript-eslint/recommended"),


vitest.config.js
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+	test: {
+		environment: 'jsdom',
+		globals: true,
```
VictoriaBeilsten-Edmands: If updating to use `vitest.setup.ts`
```suggestion
		globals: true,
		setupFiles: ['./vitest.setup.ts'],
```

VictoriaBeilsten-Edmands (changes requested):

VictoriaBeilsten-Edmands (approved):

PR #115, opened 2025-10-24
Add catalog-info.yaml

VictoriaBeilsten-Edmands:
Adds sci-react-ui to the dev portal software catalogue.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -3,7 +3,7 @@ kind: Component
 metadata:
   name: sci-react-ui
   title: Scientific React UI
-  description: Scientific React UI components, based on MUI. Early development release. A theme and component library to make websites at scientific institutions simple to create.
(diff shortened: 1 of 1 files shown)


catalog-info.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: sci-react-ui
+  title: Scientific React UI
+  description: Scientific React UI components, based on MUI. Early development release. A theme and component library to make websites at scientific institutions simple to create.
```
akademy: Maybe we should drop the early development release (are we a year old now?!)
```suggestion
  description: Scientific React UI components, based on MUI. A theme and component library to make websites at scientific institutions simple to create.
```

akademy (changes requested):

Issue #52, opened 2025-04-10
Multilevel menus


akademy:
It should be possible to create multiple menu levels (submenus) in the NavBar and the Footer.

The NavBar could dropdown on hover/select, the footer could display them beneath a heading.

akademy:
For the top, NavBar, I was imaging them to work more like your bookmark folder example. I would like them to work focus as well as hover, so that people navigating with a keyboard can still use them. MUI already has some Menu components, not sure whether you can use them here or not.

However, I'm not sold on that, so feel free to experiment if you find something better.


For the footer, I was thinking a flat appearance, no popups, just a lists, grouped together with headings. But that's a lower priority.

JamesDoingStuff:
I've put what I've done so far on the NavBar menu in #123. I still need to write some tests before it can be reviewed, but feel free to take a look it in the storybook and let me know what you think of it - happy to make changes

PR #123, opened 2025-11-17
Multi-level Navbar Menu

JamesDoingStuff:
For #52
Couple of things worth mentioning:
- I added the user-event package to make testing keyboard inputs simpler (`fireEvent` was too restrictive)
- To get the `NavMenuLink` to work with focus I had to use ref forwarding, and to support router links without duplicating code, it now uses the `NavLink`, but this meant adding ref forwarding to `NavLink` as well, lest the console fill up with React warnings.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,7 +7,7 @@ SciReactUI Changelog
 ### Added
 - New *Progress* component based on Diamond Light added.
 - New *ProgressDelayed* component so that the progress isn't shown at all when it's a small wait.
-- *NavMenu* component added for creating dropdown menus in the NavBar
+- *NavMenu* component added for creating dropdown menus in the Navbar
   - *NavMenuLink* component extends NavLink to work in the NavMenu
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


src/components/navigation/NavMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const closeMenu = () => {
+    setAnchorElement(null);
+  };
+
+  const theme = useTheme();
+
+  return (
+    <>
+      <Button
```
VictoriaBeilsten-Edmands: Can improve accessibility a bit here with [aria-haspopup](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-haspopup), [aria-expanded](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-expanded) and [aria-controls](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls) then link to the menu with `id`.

VictoriaBeilsten-Edmands (commented):
Looks good and you've handles accessibility well. Just a couple of minor points.

akademy (approved):
LG

PR #124, opened 2025-11-25
Simplify User component when we have auth

akademy:
Instead of setting each parameter individually, we can just pass in auth from useAuth.

This also adds an extra menu item to the User menu so that users can see their Keycloak profile page.


src/components/controls/User.test.tsx
```diff
@@ -127,3 +130,69 @@ describe("User", () => {
...
+  const authDummyUser = {
+    name: "",
+    givenName: "",
+    familyName: "",
+    fedId: "",
+    email: "",
+  };
+
+  it("should render", () => {
+    renderWithProviders(<User auth={authDummy} />);
```
RhysMAdey: do we need any assertions after this to check its rendered?

PR #132, opened 2026-02-04
feat: add OSV Scanner workflow

vredchenko:
## Summary

- Add automated vulnerability scanning using Google's OSV Scanner action (v2.3.2)
- Runs on PR (diff scan), push to main (full scan), daily at 03:00 UTC, and manual dispatch
- Reports findings via SARIF to GitHub Security tab

## Test plan

- [x] Verify workflow triggers on this PR
- [ ] Check Security tab for SARIF upload after merge

### DIFF
--- a/.github/workflows/osv-scanner.yml
+++ b/.github/workflows/osv-scanner.yml
@@ -1,0 +1,26 @@
+name: OSV Scanner
+
+on:
+  pull_request:
+    branches: [main]
+  merge_group:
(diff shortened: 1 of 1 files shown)


akademy (changes requested):
Now that the vulnerabilities are being highlighted, it would be a good time to fix them!

See if you can fix the current security concerns with some package updates...

PR #155, opened 2026-03-09
Add Number Input

### LINKED ISSUE #154: Add a NumberInput component
A common NumberInput (similar to VisitInput) component which can validate various number types (i.e. natural, integers, floats, scientific) and high and low limits would be useful. This is currently in use for [ViSR](https://github.com/DiamondLightSource/atlas/blob/main/apps/visr/src/components/NumberInput.tsx), but would likely enjoy use by many more in the future.

EmsArnold:
Fixes #154. 

Adds a number input component, which validates various number types, high limits, and low limits.

### DIFF
--- a/src/components/controls/NumberInput.tsx
+++ b/src/components/controls/NumberInput.tsx
@@ -104,7 +104,7 @@ const NumberInput: React.FC<NumberInputProps> = ({
   onCommit,
   commitOnReturn = true,
   commitOnBlur = true,
-  minValue = -Infinity,
+  minValue = numberMode == "natural" ? 0 : -Infinity,
   maxValue = Infinity,
 }) => {
   const [numberText, setNumberText] = useState(
--- a/src/components/controls/NumberInput.tsx
+++ b/src/components/controls/NumberInput.tsx
@@ -62,13 +62,24 @@ const NumberInputText: React.FC<NumberInputTextProps> = ({
   };
 
   const handleKeyDown = (event: { key: string }) => {
-    if (event.key === "Enter" && commitOnReturn && isValid && handleCommit) {
-      handleCommit();
+    if (
+      event.key === "Enter" &&
+      isValid &&
+      isInLimits &&
+      commitOnReturn &&
+      handleCommit
+    ) {
+      if (commitOnBlur) {
+        const input = document.activeElement as HTMLElement;
+        input.blur();
+      } else {
+        handleCommit();
+      }
     }
   };
 
   const handleBlur = () => {
-    if (isValid && commitOnBlur && handleCommit) {
+    if (isValid && isInLimits && commitOnBlur && handleCommit) {
       handleCommit();
     }
   };
--- a/src/components/controls/NumberInput.tsx
+++ b/src/components/controls/NumberInput.tsx
@@ -48,18 +48,31 @@ const NumberInputText: React.FC<NumberInputTextProps> = ({
   maxValue,
 }) => {
   const numberRegex = Modes[numberMode];
-
   const validHelperText = !helperText
     ? ""
     : `A ${numberMode} number. Limits: ${minValue} to ${maxValue}`;
 
-  const handleInputChange = (value: string) => {
+  function calculateHelperText() {
+    return !isValid || !isInLimits
+      ? !isValid
+        ? "Invalid input"
+        : "Outside limits"
+      : validHelperText;
+  }
+
+  function checkValidity(value: string) {
+    setIsValid(numberRegex.test(value));
+  }
+
+  function checkLimits(value: string) {
     setIsInLimits(
       parseFloat(value) >= minValue && parseFloat(value) <= maxValue,
     );
-    setIsValid(numberRegex.test(value));
+  }
+
+  function updateText(value: string) {
     setNumberText(value);
-  };
+  }
 
   const handleKeyDown = (event: { key: string }) => {
     if (
@@ -88,17 +101,15 @@ const NumberInputText: React.FC<NumberInputTextProps> = ({
     <TextField
       label={label}
       value={numberText}
-      onChange={(e) => handleInputChange(e.target.value)}
+      onChange={(e) => {
+        checkValidity(e.target.value);
+        checkLimits(e.target.value);
+        updateText(e.target.value);
+      }}
       onKeyDown={handleKeyDown}
       onBlur={handleBlur}
       error={!isValid || !isInLimits}
-      helperText={
-        !isValid || !isInLimits
-          ? !isValid
-            ? "Invalid input"
-            : "Outside limits"
-          : validHelperText
-      }
+      helperText={calculateHelperText()}
       variant="outlined"
     />
   );
--- a/src/components/controls/NumberInput.tsx
+++ b/src/components/controls/NumberInput.tsx
@@ -24,8 +24,8 @@ interface NumberInputTextProps {
   handleCommit?: () => void;
   commitOnReturn?: boolean;
   commitOnBlur?: boolean;
-  minValue?: number;
-  maxValue?: number;
+  minValue: number;
+  maxValue: number;
 }
 
 const NumberInputText: React.FC<NumberInputTextProps> = ({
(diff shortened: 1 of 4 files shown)


src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const handleCommit = () => {
+    const parsedValue: number = parseFloat(numberText);
+    if (onCommit) {
+      onCommit(parsedValue);
+    }
+  };
+
+  return (
+    <>
+      {
```
akademy: We don't need "<>" "{" here

src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  return (
+    <TextField
+      label={label}
+      value={numberText}
+      onChange={(e) => handleInputChange(e.target.value)}
+      onKeyDown={handleKeyDown}
+      onBlur={handleBlur}
+      error={!isValid}
+      helperText={!isValid ? "Invalid input" : helperText}
```
akademy: Can we have a different message when the values are outside the min / max value, it's just slighly easier for users to understand. "Outside of range" or " too high", something like that.

src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,142 @@
...
+  commitOnReturn = true,
+  commitOnBlur = true,
+  minValue = numberMode == "natural" ? 0 : -Infinity,
+  maxValue = Infinity,
+}) => {
+  const [numberText, setNumberText] = useState(
+    !defaultValue ? "" : defaultValue.toString(),
+  );
+  const [isValid, setIsValid] = useState(
+    !defaultValue ? true : Modes[numberMode].test(defaultValue.toString()),
```
akademy, line 136: Do we want to check the limits here too?

akademy (changes requested):
This looks great. Just a few minor suggestions.

src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,142 @@
...
+  setIsValid,
+  handleCommit,
+  commitOnReturn,
+  commitOnBlur,
+  minValue,
+  maxValue,
+}) => {
+  const numberRegex = Modes[numberMode];
+
+  const helperText = `A ${numberMode} number. Limits: ${minValue} to ${maxValue}`;
```
VictoriaBeilsten-Edmands: Can you make it optional to display this helper text?

src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  const checkLimits = (value: string) => {
+    return parseFloat(value) >= minValue && parseFloat(value) <= maxValue;
+  };
+
+  const handleInputChange = (value: string) => {
+    setIsValid(numberRegex.test(value) && checkLimits(value));
```
VictoriaBeilsten-Edmands: Could set different error message at `checkLimits` level.

src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,142 @@
...
+    }
+  };
+
+  return (
+    <TextField
+      label={label}
+      value={numberText}
+      onChange={(e) => handleInputChange(e.target.value)}
+      onKeyDown={handleKeyDown}
+      onBlur={handleBlur}
```
VictoriaBeilsten-Edmands, line 93: If you do not set either `commitOnBlur` or `commitOnReturn` to `false` and hit `return` you get in a loop. `Enter` -> `Blur`

VictoriaBeilsten-Edmands (changes requested):

src/components/controls/NumberInput.tsx
EmsArnold: Removed

src/components/controls/NumberInput.tsx
EmsArnold: Helper text for invalid input due to limits is now "Outside limits"

src/components/controls/NumberInput.tsx
EmsArnold: Limits are now checked at all relevent levels, and corresponding test has been added

src/components/controls/NumberInput.tsx
EmsArnold: Helper text is default, but can now be turned off

src/components/controls/NumberInput.tsx
EmsArnold: This loop seems to have only been an issue due to `alert` stealing focus from the original element. I've now added logic to blur the element if `commitOnBlur` is `true` (thus using the `onBlur` action to trigger the `handleCommit` when the `enter` key is pressed), otherwise `handleCommit` for `onKeyDown` happens as normal.
I tried it out a few times without managing to trigger a loop, but let me know if this doesn't work.

src/components/controls/NumberInput.tsx
akademy: This was resolved via a previous comment.

src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,173 @@
...
+  minValue,
+  maxValue,
+}) => {
+  const numberRegex = Modes[numberMode];
+
+  const validHelperText = !helperText
+    ? ""
+    : `A ${numberMode} number. Limits: ${minValue} to ${maxValue}`;
+
+  const handleInputChange = (value: string) => {
```
VictoriaBeilsten-Edmands: isValid and isInLimits are both derived from props and numberText so we can remove those two states and move their calculation out of handleInputChange. `handleInputChange` then only needs `setNumberText(value)`

src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  return (
+    <TextField
+      label={label}
+      value={numberText}
+      onChange={(e) => handleInputChange(e.target.value)}
+      onKeyDown={handleKeyDown}
+      onBlur={handleBlur}
+      error={!isValid || !isInLimits}
+      helperText={
```
VictoriaBeilsten-Edmands: Can calculate helperText outside of return

VictoriaBeilsten-Edmands (changes requested):

src/components/controls/NumberInput.tsx
EmsArnold: I believe I've addressed this comment and the one below - let me know if this is not what you meant. I think limits and validity still need to be checked on each change as they inform error state and ability to commit.

VictoriaBeilsten-Edmands: You have three states, but they're derived from the same props, so you can simplify by removing isValid and isInLimits as states, calculating these and setting to const.

src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,184 @@
...
+  commitOnReturn = true,
+  commitOnBlur = true,
+  helperText = true,
+  minValue = numberMode == "natural" ? 0 : -Infinity,
+  maxValue = Infinity,
+}) => {
+  const [numberText, setNumberText] = useState(
+    !defaultValue ? "" : defaultValue.toString(),
+  );
+  const [isValid, setIsValid] = useState(
```
VictoriaBeilsten-Edmands: Remove state with:
`const isValid = Modes[numberMode].test(numberText)`

src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,184 @@
...
+    !defaultValue ? true : Modes[numberMode].test(defaultValue.toString()),
+  );
+  const [isInLimits, setIsInLimits] = useState(
```
VictoriaBeilsten-Edmands: Remove state here as above:
`const isInLimits = .....`

src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,184 @@
...
+  };
+
+  return (
+    <TextField
+      label={label}
+      value={numberText}
+      onChange={(e) => {
+        checkValidity(e.target.value);
+        checkLimits(e.target.value);
+        updateText(e.target.value);
```
VictoriaBeilsten-Edmands:
```suggestion
      onChange={(e) => {
        setNumberText(e.target.value);
```

VictoriaBeilsten-Edmands (changes requested):
This looks good. There are two states that can be removed.

VictoriaBeilsten-Edmands (approved):

PR #160, opened 2026-04-10
Upgrade to MUI v7

gfrn:
MUI v7 (as a stepping stone to v9) has been out for a while, but v8 and v9 are not supported by JSON Forms, so I've upgraded to v7 and took care of some breaking changes.

I've also moved `material-icons` to be a base dependency, which if we use the tree-shaking feature by importing from default exports, has a minimal impact on bundle size (instead of importing all of `material-icons`, since on v7, the bundling issue that required packages to use barrel imports was fixed.

PR #161, opened 2026-04-10
Split package into modules

gfrn:
Since not everyone might use JSON Forms and all the features `sci-react-ui` has, I've made this PR as a proposal to allow code splitting of modules, that way, one can import only what they need and reduce bundle size (similar to the approach `material-ui` has, for instance). 

We also reduce the number of dependencies, as that way, one doesn't end up importing JSON-Forms or KeycloakJS if they're only using navbars.

Also, as a follow-up to #160, I've used default exports for `material-icons`, as that only imports the icons we're actually using.

I've also rewritten [package exports](https://nodejs.org/api/packages.html#package-entry-points) to use a slightly more modern (supported since Node v12) structure, where instead of using `main` and `module`, we can assign several separate exports and entry points, which makes tree-shaking and code splitting easier.

However, this is still backwards compatible, and anyone using barrel imports from `@diamondlightsource/sci-react-ui` directly should not be affected, nor should anyone building this package on Node <v12 (they shouldn't, but it still works)

This requires #160 as it builds on top of those changes.


src/components/navigation/NavMenu.tsx
```diff
@@ -8,7 +8,7 @@ import {
   type MenuItemProps,
 } from "@mui/material";
 import React, { useState, forwardRef, useId } from "react";
-import { ExpandMore as ExpandMoreIcon } from "@mui/icons-material";
+import ExpandMoreIcon from "@mui/icons-material/Expand";
```
akademy: Side note: This is why I don't like default imports. Easy to select the wrong thing without noticing.
```suggestion
import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
```

gfrn: Yeah, true, the benefits are a marginally smaller bundle size and bettertree-shaking, but JS really needs some "default name, default import function"

akademy (approved):
This looks good. So many times I've had problems with those icon imports downstream, but it looks like you may finally fixed it. Nice work.

gfrn:
> This looks good. So many times I've had problems with those icon imports downstream, but it looks like you may finally fixed it. Nice work.

I think this was down to a bug in MUI v6, at least according to the GitHub threads I found, so that's more on them than it is me

PR #162, opened 2026-04-17
Image component and magnified image

gfrn:
This adds two components we made for Evo, one that provides an "upgraded" HTML  tag that displays a placeholder if the image fails to load, and displays a loading indicator while the image loads. 

It also adds another component (mainly used for diffraction images and particle picking) that allows users to zoom in on any area in an image, with lens/zoom sizes being configurable. It also allows users to set contrast, brightness and invert colours to make features easier to detect

image

It is also built entirely with CSS, MUI and HTML, so it requires no extra dependencies.

This depends on https://github.com/DiamondLightSource/sci-react-ui/pull/161, as it would be ideal for us to only import components we're using to minimise bundle size and dependencies


package.json
```diff
@@ -52,7 +52,8 @@
     "@fontsource-variable/inter": "^5.2.8",
     "@mui/icons-material": "^7.0.0",
     "react-icons": "^5.3.0",
-    "utif": "^3.1.0"
+    "utif": "^3.1.0",
+    "@mui/icons-material": "^7.0.0"
```
VictoriaBeilsten-Edmands, line 55: This has got duplicated

VictoriaBeilsten-Edmands (commented):
Why does the zoom go outside the image when clicking on the right but inside when clicking on the left?

Image

Image

gfrn:
> Why does the zoom go outside the image when clicking on the right but inside when clicking on the left?
> Image Image

There was a 100px window where the padding didn't kick in, and it resulted in that, but I removed it because it's just better to use the native breakpoints anyway, so that should be fixed

yarn.lock
VictoriaBeilsten-Edmands, line 1: We're using pnpm. Would you take this one out please.

gfrn: Is it still there? I think I removed it in the previous commit

VictoriaBeilsten-Edmands: So you did!

VictoriaBeilsten-Edmands (dismissed):
Looks like a good addition.

gfrn:
Sorry, is it OK to merge this one?

VictoriaBeilsten-Edmands:
Yes. 👍

gfrn:
Ah yeah, I had to merge main, so I guess it needs another approval

akademy:
I'm seeing a problem in Storybook where when you use the zoom button (the one actually created BY storybook) the red box doesn't show the right position. I'm on firefox.

Is that reproducable for you?

akademy:
> I'm seeing a problem in Storybook where when you use the zoom button (the one actually created BY storybook) the red box doesn't show the right position. I'm on firefox.
> 
> Is that reproducable for you?

I don't know when it will affect the real world, so I'm happy for it to be merged.

akademy:
PS When this is in, I'll make a ~new~ beta release.

akademy (approved):

gfrn:
I can reproduce it in Storybook, but in real usage, because it's hooked to the window size, it actually responds to screen size changes/resizes/zoom, so I don't think it would affect real usage

I'll merge it in, then, thanks for having a look!

PR #166, opened 2026-04-30
Add stories for MUI Surface components

VictoriaBeilsten-Edmands:
Adds stories for MUI components Accordion, Paper, AppBar and Card.

### DIFF
--- a/.storybook/muiDocsParameters.ts
+++ b/.storybook/muiDocsParameters.ts
@@ -1,0 +1,14 @@
+export const muiDocsText = `
+  These components are Material UI (MUI) components.
+  
+  For full API documentation, examples, and usage guidelines, see:
+  https://v7.mui.com/material-ui/all-components
+        `;
+
+export const muiDocsParameters = {
+  docs: {
+    description: {
+      component: muiDocsText,
+    },
+  },
+};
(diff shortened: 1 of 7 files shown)


NKatti2011 (approved):
Looks good to me (LGTM)

akademy:
I haven't reviewed the code properly yet, but not expecting anything wrong, maybe a typo or something - will do that later today.

But a more general point is, I would like a message on the Storybook explaining these components are MUI directly and you should see the MUI website for full documentation. It could be done in the same way our Footer has text on its Storybook page. 

Also, I wonder if people will now confuse the MUI's AppBar with our one - maybe an explanation, or reference to the other should be added.

VictoriaBeilsten-Edmands:
I can add a comment in on the pages. Could link each component to documentation (e.g. [Button](https://mui.com/material-ui/react-button/)), but might be simplest to add a decorator which add the same message to all MUI stories and links to [all MUI documentation](https://mui.com/material-ui/all-components/).

I'll add something about the AppBar.

akademy:
Yeah, there's probably a smart way to do it.

We could do that further down the line though. It doesn't have to be part of the branch.

akademy:
Ah, I see you've already done it!

akademy:
This is a great piece of work. 

(One minor thing is, we've used "MUI_" in a few limited cases already, e.g. [src/components/navigation/Breadcrumbs.tsx](https://github.com/DiamondLightSource/sci-react-ui/blob/e610e20b7c521db7975f4c3a4759dff38e54c37f/src/components/navigation/Breadcrumbs.tsx#L4) but I suggest we change those files over to Just use"Mui" )

akademy (approved):

PR #170, opened 2026-05-01
Adds stories for MUI Data Display components

VictoriaBeilsten-Edmands:
Adds stories for MUI components Avatar, Badge, Chip, Divider, Icon, List, Table, Typography.


NKatti2011:
Could we have tests for these?

VictoriaBeilsten-Edmands:
> Could we have tests for these?

I haven't added tests because we are not re-exporting MUI components so any tests would be testing the underlying MUI code. These are just meant as an interactive way to look at MUI components with our theme.

akademy (approved):
These all look good.

PR #171, opened 2026-05-05
Suggested MUI component wrapper with a few examples.

akademy:
A component to wrap MUI components.

A couple of examples with Alert and Chip.

Note, that the default HTML Entity for the `forwardRef` is `HTMLDivElement` but this can be overridden with what ever you like.


VictoriaBeilsten-Edmands (approved):
Looks great. Just needs to change `any` to pass CI.

akademy:
Should be good now.

VictoriaBeilsten-Edmands:
Looks good to merge.

PR #177, opened 2026-05-08
Introduce DiamondDS theme foundation and component updates

zoharma:
Introduced the initial Diamond Design System theme foundation, including semantic colour roles, light/dark mode support, CSS variable mappings, and integration with the MUI theming architecture.

Updated component styling, states, and overrides across inputs, buttons, checkboxes, radio buttons, overlays, borders, and theme icons.


.storybook/preview.tsx
```diff
@@ -1,15 +1,53 @@
...
 const TextThemeDiamond = "Theme: Diamond";
 
+const TextThemeDiamondDS = "Theme: DiamondDS";
+
+function resolveTheme(selectedTheme: string, mode: "light" | "dark") {
+  switch (selectedTheme) {
+    case TextThemeBase:
+      return GenericTheme;
+    case TextThemeDiamondDS:
+      return mode === "dark" ? DiamondDSThemeDark : DiamondDSTheme;
```
akademy: Usually a theme should include the dark and light mode already. i.e. There just needs to be a single theme with two colour schemes inside it.

Some info here.
https://v7.mui.com/material-ui/customization/palette/#color-schemes

.storybook/preview.tsx
```diff
@@ -1,15 +1,53 @@
...
+  mode: "light" | "dark";
+  doc: Document;
+}) {
+  useLayoutEffect(() => {
+    const root = doc.documentElement; // <html>
+    root.setAttribute("data-mode", mode);
+
+    // Optional: keep class too if your CSS supports it
+    root.classList.toggle("dark", mode === "dark");
+    root.classList.toggle("light", mode === "light");
```
akademy: I'm not sure why we need data-mode and class here. It looks like all the themes already use class as they inherit from the BaseTheme.
More info here:
https://v7.mui.com/material-ui/customization/css-theme-variables/configuration/#toggling-dark-mode-manually

If this hack is needed, does it also have to be run in end user apps downstream?

tsconfig.json
```diff
@@ -1,11 +1,7 @@
 {
```
akademy, line 1: This doesn't seem to have any changes except cosmetic.

src/index.ts
```diff
@@ -25,7 +25,7 @@ export * from "./components/helpers/jsonforms";
 // themes
 export * from "./themes/BaseTheme";
 export * from "./themes/DiamondTheme";
-export * from "./themes/DiamondOldTheme";
+export * from "./themes/DiamondDSTheme";
```
akademy, line 28: Is there some reason why we've removed the old theme?

src/themes/ThemeProvider.tsx
```diff
@@ -1,26 +1,37 @@
...
+  mode?: DSMode; // 'light' | 'dark' (adding 'system' for future use)
 }
 
-const ThemeProvider = function ({
+export function ThemeProvider({
   children,
-  theme = GenericTheme,
   baseline = true,
   defaultMode = "system",
+  mode = "light", // default to light mode (for now)
```
akademy: We appear to have two modes now... 

It'd be more sensible to use the existing MUI one.

.storybook/preview.tsx
```diff
@@ -28,12 +66,16 @@ export const decorators = [
...
+    // ensure we target the preview iframe document
+    const doc: Document = context?.canvasElement?.ownerDocument ?? document;
     return (
       <ThemeProvider
-        theme={selectedTheme == TextThemeBase ? GenericTheme : DiamondTheme}
-        defaultMode={selectedThemeMode == TextLight ? "light" : "dark"}
+        theme={resolveTheme(selectedTheme, mode)}
+        defaultMode={mode}
       >
+        <ApplyModeToPreviewDoc mode={mode} doc={doc} />
```
akademy: Ins't the code in ApplyModeToPreviewDoc already embedded in ThemeProvider ? It seems to be running it twice here.

.storybook/preview.tsx
```diff
@@ -75,18 +117,6 @@ const preview: Preview = {
     },
     backgrounds: { disable: true },
     layout: "fullscreen",
-    options: {
```
akademy, line 78: Why were these removed? Is it not doing anything?

.storybook/ThemeSwapper.tsx
```diff
@@ -21,12 +21,14 @@ export const TextDark = "Mode: Dark";
...
-  //if( !mode ) return
 
   useEffect(() => {
     const selectedThemeMode = context.globals.themeMode || TextLight;
-    setMode(selectedThemeMode == TextLight ? "light" : "dark");
-  }, [context.globals.themeMode]);
+    const nextMode = selectedThemeMode === TextLight ? "light" : "dark";
+
+    setMode(nextMode);
+    document.documentElement.setAttribute("data-mode", nextMode);
```
akademy: This seems really hacky. I'd avoid directly changing anything in the document given that React is trying to control the entire app.

src/themes/DiamondDSTheme.ts
```diff
@@ -0,0 +1,1122 @@
...
+        onSolid: "var(--ds-on-info-solid)",
+
+        contrastTextChannel: "var(--ds-on-info-channel)",
+        mainChannel: "var(--ds-info-mainChannel)",
+        lightChannel: "var(--ds-info-lightChannel)",
+        darkChannel: "var(--ds-info-darkChannel)",
+      },
+
+      grey: {
+        50: "#F8F8FA",
```
akademy: Is there some reason these values aren't defined in the CSS file?

src/styles/diamondDS/diamond-ds-roles.css
```diff
@@ -0,0 +1,430 @@
...
+  --ds-focus-ring-danger: var(--ds-danger-accent);
+  --ds-focus-ring-warning: var(--ds-warning-accent);
+  --ds-focus-ring-success: var(--ds-success-accent);
+  --ds-focus-ring-info: var(--ds-info-accent);
+  --ds-focus-ring-brand: var(--ds-brand-accent);
+  --ds-focus-ring-highlight: var(--ds-highlight-accent);
+}
+
+.dark,
+html[data-mode="dark"] {
```
akademy: This should just be for the root. So something like:
```suggestion
:root[data-mode="dark"] {
```
or 
```suggestion
:root.dark{
```

akademy (commented):
I've just highlighted a few things for me to check tomorrow, just getting my head around what's there. @zoharma no need for any changes yet.

.storybook/preview.tsx
akademy: I see it does need to be run, by ApplyModeToPreviewDoc, downstream.

src/index.ts
zoharma: I don't see that it's being used by the dropdown. We can keep it...

.storybook/ThemeSwapper.tsx
zoharma: I don't think it works well without it

src/themes/DiamondDSTheme.ts
zoharma: MUI theme expects these. For now I thought to keep it simple, but a more advanced tokenised way would be to add to the CSS:
```
:root {
  --ds-grey-50: #f8f8fa;
  --ds-grey-100: #eef1f5;
  --ds-grey-200: #e6e9f0;
  --ds-grey-300: #dde1e8;
  --ds-grey-400: #bcc2cd;
  --ds-grey-500: #a5acb8;
  --ds-grey-600: #8a90a0;
  --ds-grey-700: #505563;
  --ds-grey-800: #2c3140;
  --ds-grey-900: #1a1c23;
}
```
Then in the MUI theme:

```
grey: {
  50: "var(--ds-grey-50)",
  100: "var(--ds-grey-100)",
  200: "var(--ds-grey-200)",
  300: "var(--ds-grey-300)",
  400: "var(--ds-grey-400)",
  500: "var(--ds-grey-500)",
  600: "var(--ds-grey-600)",
  700: "var(--ds-grey-700)",
  800: "var(--ds-grey-800)",
  900: "var(--ds-grey-900)",
},
```

src/styles/diamondDS/diamond-ds-roles.css
zoharma: So for this theme, we could drop the class fallback? I wasn't sure about that.

src/index.ts
akademy: It's not available in Storybook, but it can be used downstream. It's just there in case someone says "we loved the other one"! But hopefully, no one will use it!

src/styles/diamondDS/diamond-ds-roles.css
akademy: My thinking was, we should just support one. It'll just simplify everything which is worth a breaking change if need be.

src/themes/DiamondDSTheme.ts
```diff
@@ -0,0 +1,1122 @@
...
+const getOverlayInset = (token = "var(--ds-overlay-hover)") =>
+  `inset 0 0 0 9999px ${token}`;
+
+// --- Theme factory ---
+
+export const createDiamondTheme = (mode: DSMode): Theme => {
+  const DiamondDSThemeOptions = mergeThemeOptions({
+    typography: {
+      fontFamily: [
+        "Inter Variable",
```
akademy: Should this be in quotes, like "Segoe UI" is?

```suggestion
       '"Inter Variable"',
```

package.json
VictoriaBeilsten-Edmands, line 1: Are any of these version bumps needed for the theme changes in this PR?

VictoriaBeilsten-Edmands (approved):

PR #181, opened 2026-05-13
Upgrade to pnpm v10.20 & add minimumReleaseAge

VictoriaBeilsten-Edmands:
Adds minimumReleaseAge and trustPolicy for dependencies.
Upgrades to pnpm v10.26 to allow minimumReleaseAge usage.
Adds override for fast-uri v3.1.2 due to npm security audit.
Added eslint/js to deps

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
     "@jsonforms/core": "^3.7.0",
     "@jsonforms/material-renderers": "^3.7.0",
     "@jsonforms/react": "^3.7.0",
-    "@mui/icons-material": "^7.0.0",
     "@mui/material": "^7.0.0",
     "react": "^18.3.1"
   },
@@ -133,5 +132,5 @@
       "fast-uri@<3.1.2": "3.1.2"
     }
   },
-  "packageManager": "pnpm@10.20.0"
+  "packageManager": "pnpm@10.26.0"
 }
--- a/package.json
+++ b/package.json
@@ -49,10 +49,10 @@
     "storybook:publish": "gh-pages -b storybook/publish -d storybook-static"
   },
   "dependencies": {
+    "@mui/icons-material": "^7.0.0",
     "keycloak-js": "^26.2.1",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -49,19 +49,19 @@
...
+    "utif": "^3.1.0"
   },
   "peerDependencies": {
     "@emotion/react": "^11.13.3",
     "@emotion/styled": "^11.13.0",
     "@jsonforms/core": "^3.7.0",
     "@jsonforms/material-renderers": "^3.7.0",
     "@jsonforms/react": "^3.7.0",
-    "@mui/material": "^7.0.0",
     "@mui/icons-material": "^7.0.0",
```
akademy: We appear to have this both in dependencies and peerDependencies...

I was speaking to Guilherme about something related, and he said icons are fine under dependencies, as they (should) only include the icons we are using.

package.json
```diff
@@ -49,19 +49,19 @@
     "storybook:publish": "gh-pages -b storybook/publish -d storybook-static"
   },
   "dependencies": {
+    "@mui/icons-material": "^7.0.0",
     "keycloak-js": "^26.2.1",
```
akademy: Guilherme mentioned moving this to peer. I'm in two minds about it. Do you have an opinion? 

(I'm asking under this MR, because I assume it's better to move it to peer so that it is easier to overwrite downstream?)

VictoriaBeilsten-Edmands: Could move it to peerDependencies and put as an optionalDependency but would need some documentation to point out that it needs to be installed.

akademy: Yeah, lets move it out.

VictoriaBeilsten-Edmands: Ok, moved.

akademy (approved):
I wonder if we should count the keycloak change as a breaking change... but I'll approve for now.

VictoriaBeilsten-Edmands:
You're right, it should be marked as a breaking change. I'll update.

PR #189, opened 2026-05-28
Fix devcontainer workspace mounting configuration

zoharma:
Fix `workspaceMount` source so it may not get lost again


akademy:
Not sure what this is trying to do, but I don't use DevContainers

zoharma:
> Not sure what this is trying to do, but I don't use DevContainers

There was an issue that caused my container to lose sight of where the actual files were, and this was supposed to help. 

I have just migrated from Podman to Docker so maybe not needed. Let's put it on hold.

akademy:
This may be resurrected at some point, but for now, it's being closed.

PR #191, opened 2026-05-28
Add critical level to osv scanner

VictoriaBeilsten-Edmands:


.github/workflows/osv-scanner.yml
```diff
@@ -20,7 +20,13 @@ jobs:
   scan-pr:
     if: github.event_name == 'pull_request' || github.event_name == 'merge_group'
     uses: google/osv-scanner-action/.github/workflows/osv-scanner-reusable-pr.yml@v2.3.2
+    with:
+      scan-args: |
+        --min-severity=CRITICAL
```
gfrn: I believe the severities are score-based rather than name based, so it'd be


```suggestion
        --min-severity=9
```

gfrn:
Looks good to me, if it keeps giving us trouble, we can maybe move it to an "advisory" role, as not all critical vulns are exploitable, so if it starts blocking too much, we can leave it up to us to decide what's worth blocking

gfrn (approved):

akademy (approved):

PR #195, opened 2026-05-29
Fixed border issue on disabled state and added override for read-only

zoharma:


VictoriaBeilsten-Edmands:
Looks good. Will approve once other changes are in.

VictoriaBeilsten-Edmands (approved):

PR #205, opened 2026-06-04
Fix for dash appearing when the number zero should display.

akademy:
Fix a value appearing as dash "-" when the value should display as zero.

### DIFF
--- a/src/components/helpers/jsonforms/components/DataBox.test.tsx
+++ b/src/components/helpers/jsonforms/components/DataBox.test.tsx
@@ -16,6 +16,33 @@ describe("DataBox", () => {
     expect(container.querySelector("dd span.empty")).not.toBeInTheDocument();
   });
 
+  test("Should render data if zero and label", () => {
+    const data = "0";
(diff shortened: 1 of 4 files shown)


src/components/helpers/jsonforms/components/DataBox.test.tsx
```diff
@@ -16,15 +16,42 @@ describe("DataBox", () => {
...
+  test("Should render data if zero number and label", () => {
+    const data = 0;
+    const { container } = render(<DataBox {...props} data={data} />);
+
+    expect(screen.getByText(props.label)).toBeInTheDocument();
+    expect(screen.getByText(data)).toBeInTheDocument();
+    expect(container.querySelector("dd span.empty")).not.toBeInTheDocument();
+  });
+
+  test("Should render data if zero number and label", () => {
```
RhysMAdey: very small but should this and the above have different test names?

akademy: Fair enough. I've changed the names.

RhysMAdey (approved):
Looks good to me

PR #207, opened 2026-06-05
Update components to new DS theme

VictoriaBeilsten-Edmands:


src/components/navigation/Breadcrumbs.tsx
```diff
@@ -73,8 +61,8 @@ const Breadcrumbs = ({
   const crumbs: CustomLink[] = getCrumbs(path);
 
   return (
-    <BarStyled {...props}>
-      <MuiBreadcrumbsStyled
+    <Bar color="primary" variant="subtle" {...props}>
```
zoharma: I would suggest to make it as paper and elevation, or maybe easiest with `surface.subtle`

zoharma: Same for footer

src/components/navigation/Footer.tsx
```diff
@@ -62,35 +48,23 @@ const FooterLink = ({
...
       {children}
     </Link>
   );
 };
 
-const BarStyled = styled(Bar)<BarSlotsProps>(({ theme }) => ({
-  position: "relative",
-  bottom: 0,
-  marginTop: "auto",
-  backgroundColor: theme.vars.palette.primary.main,
```
zoharma, line 91: I would suggest to make it as paper and elevation, or maybe easiest with `surface.subtle`

src/components/controls/Bar.stories.tsx
```diff
@@ -10,24 +11,58 @@ const meta: Meta<typeof Bar> = {
...
-    }}
-  >
-    {children}
-  </div>
-);
+export const Default: Story = {
+  args: {
+    leftSlot: <Typography variant="body1">Left</Typography>,
+    centreSlot: <Typography variant="body1">Centre</Typography>,
+    rightSlot: <Typography variant="body1">Right</Typography>,
```
akademy: The removed `Slot` component was to indicate the area where the slots where. But maybe we could just indicate this with more specific text.

```suggestion
    leftSlot: <Typography variant="body1">Left & Children Slot</Typography>,
    centreSlot: <Typography variant="body1">Centre Slot</Typography>,
    rightSlot: <Typography variant="body1">Right Slot</Typography>,
```

src/components/controls/Bar.test.tsx
```diff
@@ -38,7 +38,7 @@ describe("Bar", () => {
     // check new style is set
     expect(headerComputedStyle.border).toBe(borderStyle);
     // Check default values are still set
-    expect(headerComputedStyle.height).toBe("auto");
+    expect(headerComputedStyle.height).not.toBe("0px");
```
akademy, line 41: This isn't checking what the default is. Only what it isn't (and it could be many things)

May as well take it out, unless we know what a default should be.

src/public/diamond/logo-dark-surface.svg
akademy, line 1: What is the significance of the name change?

src/components/navigation/Footer.tsx
```diff
@@ -51,8 +39,6 @@ const FooterLink = ({
   href,
   ...props
```
akademy: `props` isn't used. Should it be?

src/components/controls/Bar.tsx
```diff
@@ -8,35 +8,30 @@ import {
...
 } from "@mui/material";
+import { Theme } from "@mui/material/styles";
+
+type IntentColour =
+  | "primary"
+  | "secondary"
+  | "error"
+  | "warning"
+  | "info"
+  | "success";
```
akademy: A "brand" intent would work really well here.

src/components/navigation/Breadcrumbs.tsx
VictoriaBeilsten-Edmands: I've changed it to use `subtle` variant of `Bar` without `color` defined which uses `surface.subtle` and `text.primary`

src/components/controls/ColourSchemeButton.tsx
```diff
@@ -1,47 +1,34 @@
...
-          color: "yellow",
-        },
-      }}
-      aria-label={`Colour scheme switcher: ${colourScheme}`}
+      aria-label={`Colour scheme switcher: ${mode}`}
       {...props}
       onClick={(event) => {
-        setColourScheme(isDark() ? ColourSchemes.Light : ColourSchemes.Dark);
-        if (props.onClick) props.onClick(event);
+        setMode?.(isDark ? "light" : "dark");
```
akademy, line 15: When is `setMode` not defined? Also is `mode` also undefined? What happens then?

src/components/controls/Logo.tsx
```diff
@@ -1,26 +1,60 @@
...
 
   return (
-    logo && (
-      <ImageColourSchemeSwitch
-        image={logo}
-        interchange={interchange}
-        style={style}
-      />
-    )
+    <img
```
akademy: Why is this nolonger using `ImageColourSchemeSwitch` ?

src/components/controls/ScrollableImages.stories.tsx
```diff
@@ -73,13 +73,6 @@ export const NoNumbers: Story = {
   },
 };
 
-export const DifferentBackgroundColour: Story = {
```
akademy, line 76: Are we pretending you can't change the background colour or can you actually not do this now?

akademy (changes requested):
I've added a few things. Also got the following questions (not about specific lines/files)
- AppTitleBar, no longer matches the NavBar colours. Is there some way to set it to the correct colours and add it as a Storybook example?
- Footer seems to have an odd background colour now. (You can see original matched the NavBar too https://diamondlightsource.github.io/sci-react-ui/?path=/docs/components-navigation-footer--docs )
- The ColourSchemeButton no longer looks cool. All the hover colours have gone. This is a sad day.
- The ColourSchemeButton also no longer matches the styling of the User control (i.e. with a filled background)

src/public/diamond/logo-dark-surface.svg
VictoriaBeilsten-Edmands: To make it less confusing what "dark" means - is it dark text in the logo, or dark theme? Maybe a better name would be logo-on-dark-surface and logo-on-light-surface

src/components/controls/ScrollableImages.stories.tsx
VictoriaBeilsten-Edmands: Just pretending

src/components/controls/Bar.tsx
zoharma: Brand is not `IntentColour`, but rather it's identity. 

Actually, I don't think we need status-intent, or even any intent to show here, only backgrounds and brand

@VictoriaBeilsten-Edmands

src/components/controls/Logo.tsx
VictoriaBeilsten-Edmands: I was dealing with the issue of brand-coloured background which is dark in both light/dark mode. I should be able to update ImageColourSchemeSwitch to deal with this though.

src/components/controls/ColourSchemeButton.tsx
VictoriaBeilsten-Edmands: Before the colour scheme context is fully initialized.
https://mui.com/material-ui/customization/css-theme-variables/configuration/

src/components/navigation/Footer.tsx
VictoriaBeilsten-Edmands: Removed

src/components/controls/Bar.tsx
VictoriaBeilsten-Edmands: Updated to keep primary and secondary but remove other intents.

src/components/controls/Logo.tsx
VictoriaBeilsten-Edmands: This is using `ImageColourSchemeSwitch` again.

VictoriaBeilsten-Edmands:
AppTitleBar, Footer, Breadcrumbs and Navbar all use the Bar component and can use all the colours (variant + surface combinations) that are available in Bar. They have their own defaults. I've added some stories for these.
The ColourSchemeButton and User Avatar styling are similar now. The hover colours are still gone though. @Zohar was talking about potentially having a 'highlight' colour.

The tests have been updated to pass.

zoharma (dismissed):

src/components/controls/AppTitlebar.stories.tsx
```diff
@@ -15,10 +15,25 @@ const meta: Meta<typeof AppTitlebar> = {
...
+    variant: { table: { disable: true } },
+    elevation: { table: { disable: true } },
+
+    leftSlot: { control: false },
+    centreSlot: { control: false },
+    rightSlot: { control: false },
+    children: { control: false },
+  },
+
+  args: {
```
douglaswinter: I think you should remove this args from here. It's explicitly defined in Simple, and it appears in stories where it's not set.

src/components/controls/ScrollableImages.stories.tsx
douglaswinter: Maintaining the intent of this PR you could keep this story but change the bg colour to something from the palette e.g. "primary.container"

src/components/controls/Logo.stories.tsx
```diff
@@ -41,7 +41,7 @@ export const TheShortLogo: Story = {
 
 export const LightLogoForDarkTheme: Story = {
```
douglaswinter, line 64: We discussed in person that there are two different behaviours that are probably worth supporting: 1) Inverting the sources to get light image on dark mode and vice-versa, and 2) fixing the logo source regardless of colour scheme. I would suggest making a separate issue for this.

src/themes/DiamondDSTheme.ts
douglaswinter, line 1: Might be controversial at this point, but what do you think about renaming this to DiamondDesignSystem?

douglaswinter (commented):
The components look great across light/dark in my opinion. Just a couple of comments.

douglaswinter (dismissed):
Looks very good.

PR #209, opened 2026-06-09
Introduce Lucide based Icon component and updates

zoharma:
- Added Lucide Icons as basis for all icons
- Updated components, stories with relevancy to icons 
- Updated MuiWrapped
- Updated and removed dependency for `react-icons`
- Updated and removed dependency for `@mui/icons-material`
- Removed `react-icons` package

_Note: We can also safely remove `@mui/icons-material` package but not sure of others may use it_

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   "dependencies": {
     "@fontsource-variable/inter": "^5.2.8",
     "@mui/icons-material": "^7.0.0",
-    "react-icons": "^5.3.0",
+    "lucide-react": "^1.17.0",
     "utif": "^3.1.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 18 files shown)


akademy:
I was expecting this to just replace the icons line by line. Can you explain what the additional complexity is trying to achieve? (in particular, all the files in DataDisplay/Icons)

zoharma:
> I was expecting this to just replace the icons line by line. Can you explain what the additional complexity is trying to achieve? (in particular, all the files in DataDisplay/Icons)

As far as I understand, it's better to run Lucide Icons through `MuiSvg` as then props can be passed through to it. Also, the sizes (expanded list) and weight (Lucide specific) are not directly refelcted, so this covers for that.

Key icons will be in the theme, but if an app wants to use other icons they can import it from Lucide, and then use `LucideIcon` as the component with the right size as needed. We need to document that potentially.

PR #212, opened 2026-06-11
Add onBlur submit option to VisitInput

### LINKED ISSUE #153: Add option to submitOnBlur for VisitInput
"VisitInput" currently has the option for a submission button, or submit on return. However, there is no obvious cue for a user that a valid value has not been submitted until either of these actions happens.

Adding an option to "submitOnBlur" by using the current "onSubmit" functionality and the "onBlur" option available within mui's TextField would make this component a bit more user-friendly.

VictoriaBeilsten-Edmands:
Closes #153

### DIFF
--- a/src/components/controls/VisitInput.stories.tsx
+++ b/src/components/controls/VisitInput.stories.tsx
@@ -34,6 +34,15 @@ export const InitialVisitWithSubmitButton: Story = {
   },
 };
 
+export const InitialVisitWithSubmitOnBlur: Story = {
+  args: {
+    onSubmit: handleSubmit,
+    visit: { proposalCode: "xx", proposalNumber: 99999, number: 7 },
+    submitButton: false,
+    submitOnBlur: true,
+  },
+};
+
 export const InitialVisitWithoutReturnKeySubmission: Story = {
   args: {
     onSubmit: handleSubmit,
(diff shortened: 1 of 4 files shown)


akademy:
This is being reviewed... behind the scenes...

src/components/controls/VisitInput.stories.tsx
```diff
@@ -34,6 +34,15 @@ export const InitialVisitWithSubmitButton: Story = {
   },
 };
 
+export const InitialVisitWithSubmitOnBlur: Story = {
```
EmsArnold, line 37: This is [giving me the same loop that I fell into with the NumberInput](https://github.com/DiamondLightSource/sci-react-ui/pull/155#discussion_r2940574118). With both `submitOnReturn` and `submitOnBlur` set to `true`, when you hit `return` you enter a loop.

VictoriaBeilsten-Edmands: Good catch. I've got it to blur on enter key down in this case.

src/components/controls/VisitInput.tsx
```diff
@@ -64,6 +78,7 @@ const VisitInput: React.FC<VisitInputProps> = ({
   parameters,
   submitButton = true,
   submitOnReturn = true,
+  submitOnBlur = false,
 }) => {
   const [visitText, setVisitText] = useState(visitToText(visit));
   const [isValid, setIsValid] = useState(true);
```
EmsArnold: Could `isValid`/`setIsValid` be simplified in a similar manner to what you've requested for the NumberInput component?

EmsArnold (approved):

PR #216, opened 2026-06-15
Add fix for positioning of username and fedid.

akademy:
Just a quick fix, so we can do a new build.

### DIFF
--- a/src/components/controls/User.tsx
+++ b/src/components/controls/User.tsx
@@ -99,7 +99,8 @@ const User = ({
               <Box
                 sx={{
                   padding: "5px",
-                  display: { xs: "none", sm: "block" },
+                  display: { xs: "none", sm: "flex" },
+                  flexDirection: "column",
                 }}
               >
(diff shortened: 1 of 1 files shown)


gfrn (approved):

src/components/controls/User.tsx
```diff
@@ -99,7 +99,8 @@ const User = ({
               <Box
                 sx={{
                   padding: "5px",
-                  display: { xs: "none", sm: "block" },
+                  display: { xs: "none", sm: "flex" },
+                  flexDirection: "column"
```
gfrn:
```suggestion
                  flexDirection: "column",
```

To make prettier happy

gfrn:
All looks good! I just added a suggestion to get rid of the Prettier error

src/components/controls/User.tsx
akademy: Yeah, thanks. (Two line change and still broke it!)

PR #220, opened 2026-06-16
Docs: Adding foundation section - Colours

zoharma:
Introduced Foundation section with:
 1. Colour
  1.1 Theme Colours 

Changes:
- Adding Foundation section with colours page
- Adding theme colours page
- Added views for both light and dark modes
- Adding the relevant css and ability to do so.
- Remove typography and icons (to be pushed to another PR)
- **Remove `:root` from `data-mode="{mode}"` so that the mode can be used in any container**


NKatti2011 (dismissed):
LGTM

NKatti2011 (dismissed):

NKatti2011:
Just realised:

For this ticket - now that data-mode is supported on any container rather than just :root, worth flagging: ThemeProvider's flicker-suppression MutationObserver on [251](https://github.com/DiamondLightSource/sci-react-ui/pull/251) only watches document.documentElement, so nested containers using their own data-mode won't get transition suppression. Is this expected/acceptable?

zoharma:
@NKatti2011 that's a good point and it's worth exploring. The main reason for this change here is so that we can show dark mode elements in the docs in Storybook (set in light mode). There might be other usages for it in the future, for components to have their own dark/light mode settings (such as code-views and editors), but that may require further exploration.

zoharma:
@NKatti2011 Have added it as issue #258. Could we approve this?

NKatti2011 (dismissed):
Am approving but as discussed the removal of data-mode from root may cause downstream issues re flicker suppression in another PR.

NKatti2011 (dismissed):

NKatti2011 (approved):

PR #227, opened 2026-06-22
Add interactivity to ColourSchemeButton stories & remove system theme resolution default from Storybook

VictoriaBeilsten-Edmands:
Adds colour scheme switching to ColourSchemeButton and Navbar stories plus "fixed" light/dark stories by adding disable ThemeSwapper param for stories.
In Storybook: only call setMode in stories if theme has changed; remove default mode resolution.
Import useColorScheme from "@mui/material/styles".


douglaswinter (approved):
Looks good. Storybook stories work well; tested in i15-1 app and it all works as expected.

PR #228, opened 2026-06-22
Add dev mode for previewing components

VictoriaBeilsten-Edmands:


akademy:
This looks great, but wondering if we reference the package, rather than the individual components.

I add this to package.json

    "@diamondlightsource/sci-react-ui": "file://../",

Then reference the components as usual.

    import { Bar } from '@diamondlightsource/sci-react-ui';`

which seemed to work, but I'm getting odd typescript errors for the components.

akademy:
Actually, it's not a big deal. Maybe we can look into it some other time.

I'll approve.

akademy (dismissed):

akademy (approved):
Looks good. A future update should include a kitchen sink...!

PR #232, opened 2026-06-23
Import inter font into DiamondDSTheme and remove extra dep

VictoriaBeilsten-Edmands:


zoharma:
@VictoriaBeilsten-Edmands based on fontsource we may need the css extension, and also maybe italics too
```import "@fontsource-variable/inter/wght.css";```
```import "@fontsource-variable/inter/wght-italic.css";```

VictoriaBeilsten-Edmands:
Standard weights should be included, but I've added italics.

douglaswinter (approved):

PR #235, opened 2026-06-23
Add InitColorSchemeScript guidance to docs

VictoriaBeilsten-Edmands:


src/storybook/Installation.mdx
```diff
@@ -150,4 +150,32 @@ import { Meta } from "@storybook/blocks";
...
+    This ensures the correct colour scheme is applied immediately, based on system preference or saved user choice.
+
+
+      ```tsx filename="main.tsx"
+      import { InitColorSchemeScript } from "@mui/material";
+
+      function Root() {
+        return (
+          <>
+            <InitColorSchemeScript />
```
akademy: We customized the CSS tag, so we'll need something like this (i think):

```suggestion
<InitColorSchemeScript attribute="[data-mode='%s']" />
```

src/storybook/Installation.mdx
```diff
@@ -150,4 +150,32 @@ import { Meta } from "@storybook/blocks";
...
+            <App />
+          </>
+        );
+      }
+      ```
+
+      If your app renders into `index.html`, place the script at the top of the body so the mode is set before React mounts.
+
+      This helps avoid the brief light-mode flash when the user prefers dark mode.
+    </div>
```
akademy, line 181: Might be useful to link back to the MUI page too.

```suggestion

    For more info see MUI at https://v7.mui.com/material-ui/react-init-color-scheme-script/
                </div>
```

akademy (changes requested):

PR #239, opened 2026-06-24
Move font imports to index.ts

VictoriaBeilsten-Edmands:


src/index.ts
```diff
@@ -1,3 +1,7 @@
+// Fonts
+import "@fontsource-variable/inter";
+import "@fontsource-variable/inter/wght-italic.css";
```
zoharma, line 3: Need to remove the italic line. It's included in the Inter package, and this one is only in case italic is needed by itself. My mistake, sorry about that.

PR #245, opened 2026-07-01
Improve theme architecture and MRT compatibility

zoharma:
- Working towards MRT support with the new theme
- Added fallback light/dark hex values for MUI compatibility
- Updated and cleaned up theme overrides to consume design tokens only
- Reorganised the theme file and improved comments


src/themes/DiamondDSIntegrations.ts
VictoriaBeilsten-Edmands, line 1: Is this part needed to fix the error or only to apply extra styling we control? If it's not needed to fix the bug, can you put it into a separate PR, please?

zoharma: Have done so in #249

NKatti2011 (dismissed):
LGTM

src/themes/DiamondDSTheme.ts
```diff
@@ -42,8 +41,6 @@ import type { ImageColourSchemeSwitchType } from "components/controls/ImageColou
 
 /**
  * Standard argument shape for MUI style override callbacks.
- *
```
VictoriaBeilsten-Edmands, line 45: Why was this deleted?

VictoriaBeilsten-Edmands (dismissed):
The tests can be extended now to check that the vales introduced to the theme file match those in the css file.

src/themes/DiamondDSTheme.ts
zoharma: I changed it because the comment describes the whole type, not just the ownerState field. The old comment only explained one part of it, while the new one explains what the type is used for.

gfrn:
I've read through it on Friday, and I tested the new additions locally, so looks good to me

gfrn (approved):

PR #246, opened 2026-07-02
Fixed to remove flashes on load and on change of dark/light

zoharma:
Found there was flashing onLoad and on change light/dark mode in MRT.

Here are a number of fixes:

- Followed instructions in Installation guide.
- Found that adding initialisation in HTML helps
- updated button to reduce flashing on both onLoad and reload (and added some internal CSS to the button)


zoharma:
I've streamlined this but there are some strange issues in the checks and tests @VictoriaBeilsten-Edmands @akademy

VictoriaBeilsten-Edmands:
See #250

PR #249, opened 2026-07-02
Added an Integrations page for global MRT theme overrides

zoharma:
Added an Integrations page for global MRT theme overrides.

Use as follows:
---

For app intergation:
Use 
```
import {
  ThemeProvider,
  DiamondDSIntegrations,
  DiamondDSTheme,
} from "@diamondlightsource/sci-react-ui"
```

and inside `const table = useMaterialReactTable({`
```  
    ...DiamondDSIntegrations.mrtOptions(),
    mrtTheme: DiamondDSIntegrations.mrt(),
```

p.s. 
Have not updated ?path=/docs/installation--docs with the guidance as not sure of the best way to do that. It's only when using MRT. Also guidance to pull artefacts of DiamondDSIntegrations may update as we progress...


src/themes/DiamondDSIntegrations.ts
```diff
@@ -0,0 +1,35 @@
+export const DiamondDSIntegrations = {
+  /* Material React Table (MRT) */
+  mrt() {
```
VictoriaBeilsten-Edmands: Do these need to be functions or is it because of the future options?

src/themes/DiamondDSIntegrations.ts
```diff
@@ -0,0 +1,35 @@
...
+      muiTableHeadCellProps: {
+        sx: {
+          backgroundColor: "var(--ds-surface-container)",
+        },
+      },
+      muiTablePaperProps: {
+        elevation: 0,
+        sx: {
+          boxShadow: "none",
+          borderRadius: 1,
```
VictoriaBeilsten-Edmands: Does this want hardcoding?

src/themes/DiamondDSIntegrations.ts
```diff
@@ -0,0 +1,35 @@
+export const DiamondDSIntegrations = {
+  /* Material React Table (MRT) */
+  mrt() {
+    return {
+      baseBackgroundColor: "rgb(var(--ds-surface-channel))",
+      draggingBorderColor: "var(--ds-primary)",
+      // matchHighlightColor: "rgb(var(--ds-warning-mainChannel) / 0.30)",
+      menuBackgroundColor: "var(--ds-surface)",
+      pinnedRowBackgroundColor: "rgb(var(--ds-primary-mainChannel) / 0.10)",
+      selectedRowBackgroundColor: "rgb(var(--ds-primary-mainChannel) / 0.20)",
```
VictoriaBeilsten-Edmands: We don't have `--ds-primary-mainChannel`
Should be `--ds-primary-channel`?

src/themes/DiamondDSIntegrations.ts
VictoriaBeilsten-Edmands: Can mrt be renamed to mrtTheme

VictoriaBeilsten-Edmands (changes requested):
It would be useful to have a mrt story and something in the dev app using the integration.

src/themes/DiamondDSIntegrations.ts
zoharma: Done

src/themes/DiamondDSIntegrations.ts
zoharma: What do you mean?

src/themes/DiamondDSIntegrations.ts
zoharma: Updated to object

zoharma:
@VictoriaBeilsten-Edmands Quite a few updates:

- `mrtTheme` is now a plain object instead of a zero-arg function
- `DiamondDSIntegrations.mrtOptions()` now matches MUI Table's hover/selected colours, fixes header checkbox alignment, and avoids a crash from MRT's alpha() running on a var(...) colour
- `MuiTableContainer` theme override gets a shared border/radius, with the last-row divider fix so it doesn't double up
- Add `fullWidth` option to drop border/radius on full-bleed tables
- Add MRT demo (selection, hover) to dev app; 
- Document usage in MDX guide
- Add test coverage for both

src/themes/DiamondDSIntegrations.ts
VictoriaBeilsten-Edmands: Do we want this to be set to 1 like this or linked to a token in the theme instead?

zoharma: This was changed in the new update to be more rounded actually. We haven't done tokens for radius yet (beyond the default MUI) and we should do as a separate strand, spacing too.

dev/package.json
```diff
@@ -3,12 +3,11 @@
     "private": true,
     "type": "module",
     "scripts": {
-      "dev": "vite",
+      "dev": "vite --port 3000",
```
VictoriaBeilsten-Edmands, line 6: Why this change?

zoharma: Because if using both Dev site and Storybook from the same build it was clashing the ports

zoharma: Oh, and this one in particular was for the install and auto-deploy.

zoharma:
Added change request #280 for making this apply globally rather than requiring opt-in per table.

PR #252, opened 2026-07-06
feat: add tests for DiamondDSTheme & update theme to pass tests

VictoriaBeilsten-Edmands:
This PR
Adds tests for the the DiamondDSTheme and checks tokens, mappings and overrides.
Updates tokens to pass tests to synchronise channel and hex values & remove border-subtle refs.
Add onSolid to TypeTextOptions.

### DIFF
--- a/src/styles/diamondDS/DiamondDSTokens.css
+++ b/src/styles/diamondDS/DiamondDSTokens.css
@@ -199,7 +199,7 @@
 
   --ds-on-success-channel: 255 255 255;
   --ds-success-mainChannel: 24 122 47;
-  --ds-success-lightChannel: 47 154 73;
+  --ds-success-lightChannel: 47 179 68;
   --ds-success-darkChannel: 20 97 37;
 
   /* Info (Blue) */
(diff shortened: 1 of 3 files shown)


src/styles/diamondDS/DiamondDSTokens.css
```diff
@@ -324,8 +324,8 @@
 
   --ds-on-primary-channel: 11 22 56;
   --ds-primary-mainChannel: 138 167 255;
-  --ds-primary-lightChannel: 196 212 255;
-  --ds-primary-darkChannel: 165 188 255;
+  --ds-primary-lightChannel: 165 188 255;
```
zoharma: I don't mind doing it myself here or another branch, but these tokens should have the right semantic reference rather than light/dark here and in the theme file

src/themes/DiamondDSTheme.test.ts
```diff
@@ -0,0 +1,1206 @@
...
+  it("configures the DiamondDS font stack", () => {
+    expect(DiamondDSTheme.typography.fontFamily).toBe(
+      [
+        '"Inter Variable"',
+        "Inter",
+        "system-ui",
+        "-apple-system",
+        '"Segoe UI"',
+        "Roboto",
+        "Helvetica",
```
zoharma, line 196: I don't think we need all these fall backs to be absolutes, the main one is really the desired font, and serif, sans-serif, mono as applicable.

zoharma (dismissed):

src/styles/diamondDS/DiamondDSTokens.css
zoharma: because the tests rely on this I will update this here

zoharma:
@VictoriaBeilsten-Edmands et al, added a fix to switch the channel token to the more semantic representation

Issue #210, opened 2026-06-09
New component: sidebar navigation


douglaswinter:
# Change Proposal: Sidebar navigation component

## What is being proposed?

A new component representing main navigation controls via a drawer; see [Figma design](https://www.figma.com/design/3313bqDfu7fEa3bzJCvEQZ/Diamond-MUI-Component-Library?node-id=11545-165646).

This was approved as part of the [design for the i15-1 'Waffle' vertical slice project](https://www.figma.com/design/bWGBDTZ9e5Ui2tQqD67lHd/Waffle-I15-1-Mail-in-experiments?node-id=3125-31205&t=hDoapX2mPZcCQgHI-0).

## Why is this needed?
It scales better than the Navbar for larger navigation structures. 

## What will change?
A new component would be introduced without affecting the others. 

## Breaking change?

- [ ] Yes
- [x] No

## Next steps
A maintainer will review this issue.
If accepted, it will be marked as `accepted` and a PR may then be opened.

akademy:
Just requesting access to the figma files

VictoriaBeilsten-Edmands:
@akademy You should have access now.

VictoriaBeilsten-Edmands:
Do the panels have a set width or is it configurable (via code)?

douglaswinter:
From what I gathered from the design, the widths are fixed, but @zoharma can comment further. I have an implementation in Atlas which I'm hoping to bring in to this library. I will create a draft PR so you can see what I have at the moment. It does not implement the expandable menu item, as we didn't need it for i15-1.

zoharma:
1. Regarding "expandable menu item" it came with the library, but I don't wish to have it as part of the top level side bar nav. We might introduce it for secondary navigation side bar if needed
2. Similarly the width of the nav should be fixed so all apperances would be similar. We could introduce alternative sizes (wider 280px, narrower 220px, but don't think its needed now
@douglaswinter @VictoriaBeilsten-Edmands

akademy:
Can we get a mockup of how this looks on a mobile (non-wide) screen.

akademy:
A few thoughts.

- Should the hoverover when minimised only show a single tooltip (as shown in figma), or should it expand the whole list, extended right, over the top of what is there? That might aid navigation (as you can read everything) and make it easier to scroll through them when there are more items than can fit on a single screen.
- From a code view, is the secondary panel just the same as the primary panel? Panels could just be added when needed.
[...]

PR #262, opened 2026-07-17
New component: Secondary Nav (Sidebar expansion)

zoharma:
This proposed changes rely on #210. 

Added a **SecondaryNav** panel that sits alongside **SidebarNav**, with support for a title, search, and grouped or expandable lists. Also added **NavigationLayout**, which combines both navigation panels and handles responsive behaviour automatically, showing them side by side on desktop and switching between them on mobile.

The secondary navigation uses compact rows and the same rounded selection style as the primary sidebar. Shared navigation types have been moved to `navigation/types.ts`.

Also added history for back button

### DIFF
--- a/src/components/navigation/NavigationLayout.stories.tsx
+++ b/src/components/navigation/NavigationLayout.stories.tsx
@@ -1,0 +1,230 @@
+import { Abc, ArrowForward, GraphicEq, Menu } from "@mui/icons-material";
+import { NavigationLayout } from "./NavigationLayout";
+import { Meta, StoryObj } from "@storybook/react";
+import React from "react";
+import {
+  AppBar,
+  Box,
+  Divider,
+  IconButton,
+  Toolbar,
+  Typography,
+} from "../MUI/MuiWrapped";
+import { Theme } from "@mui/material/styles";
+import { Logo } from "../controls/Logo";
+import { ColourSchemeButton } from "../controls/ColourSchemeButton";
+import { NavLink, MemoryRouter, type NavLinkProps } from "react-router-dom";
+
+const meta: Meta<typeof NavigationLayout> = {
+  title: "Components/Navigation/NavigationLayout",
+  component: NavigationLayout,
+  decorators: [
+    (Story) => (
+      <MemoryRouter>
+        <Story />
+      </MemoryRouter>
+    ),
+  ],
+  tags: ["autodocs"],
+  parameters: {
+    // NavigationLayout always renders SidebarNav, which is position:fixed on
+    // desktop - the story canvas's default padding wrapper would otherwise
+    // misalign it against the normal-flow SecondaryNav/main content beside it.
+    fullBleed: true,
+    docs: {
+      description: {
+        component: `Composes SidebarNav and SecondaryNav, owning the responsive coordination between them. On mobile only one drawer is visible at a time - opening the secondary panel drills in and hides the primary sidebar, and a back affordance drills back out. On desktop both panels are shown side by side. Which primary item a secondary panel belongs to (e.g. "Setup" having its own sub-navigation) is entirely up to the consumer - NavigationLayout only owns the responsive mechanics, not when the panel opens.`,
+      },
+    },
+  },
+};
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+const setupGroups = [
+  {
+    items: [
+      {
+        id: "general",
+        label: "General",
+        linkProps: { to: "/setup/general", component: NavLink },
+      },
+      {
+        id: "devices",
+        label: "Devices",
+        linkProps: { to: "/setup/devices", component: NavLink },
+      },
+      {
+        id: "permissions",
+        label: "Permissions",
+        linkProps: { to: "/setup/permissions", component: NavLink },
+      },
+    ],
+  },
+];
+
+export const WithAppBar: Story = {
+  render: () => {
+    const [sidebarOpen, setSidebarOpen] = React.useState(true);
+    const [secondaryNavOpen, setSecondaryNavOpen] = React.useState(false);
+
+    // Only "Setup" has an associated secondary panel, so its link opens it and
+    // every other top-level link closes it - in a real app this would instead
+    // be derived from the current route, not from click handlers on each link.
+    const SetupLink = React.useMemo(() => {
+      const Component = React.forwardRef<HTMLAnchorElement, NavLinkProps>(
+        (props, ref) => (
+          <NavLink
+            ref={ref}
+            {...props}
+            onClick={(e) => {
+              props.onClick?.(e);
+              setSecondaryNavOpen(true);
+            }}
+          />
+        ),
+      );
+      Component.displayName = "SetupLink";
+      return Component;
+    }, []);
+    const OtherLink = React.useMemo(() => {
+      const Component = React.forwardRef<HTMLAnchorElement, NavLinkProps>(
+        (props, ref) => (
+          <NavLink
+            ref={ref}
+            {...props}
+            onClick={(e) => {
+              props.onClick?.(e);
+              setSecondaryNavOpen(false);
+            }}
+          />
+        ),
+      );
+      Component.displayName = "OtherLink";
+      return Component;
+    }, []);
+
+    const navigation = [
+      {
+        navItems: [
+          {
+            label: "Setup",
+            icon: <Abc />,
+            linkProps: { to: "/1", component: SetupLink },
+          },
+          {
+            label: "Acquisition",
+            icon: <ArrowForward />,
+            linkProps: { to: "/2", component: OtherLink },
+            selected: true,
+          },
+          {
+            label: "Analysis",
+            icon: <GraphicEq />,
+            linkProps: { to: "/3", component: OtherLink },
+          },
+        ],
+      },
+    ];
+
+    return (
+      <Box sx={{ display: "flex" }}>
+        <AppBar
+          position="fixed"
+          color="inherit"
+          sx={{
+            zIndex: (theme: Theme) => theme.zIndex.drawer + 1,
+            borderBottom: "1px solid",
+            borderColor: "divider",
+          }}
+          elevation={0}
+        >
+          <Toolbar>
+            <IconButton
+              size="large"
+              edge="start"
+              color="inherit"
+              aria-label="menu"
+              sx={{ mr: 2 }}
+              onClick={() => setSidebarOpen(!sidebarOpen)}
+            >
+              <Menu />
+            </IconButton>
+
+            <Box sx={{ mr: 2, mt: 1.5 }}>
+              <Logo sx={{ display: "block" }} />
+            </Box>
+
+            <Divider orientation="vertical" variant="middle" flexItem />
+
+            <Typography
+              variant="h6"
+              noWrap
+              component="div"
+              sx={{ ml: 1.5, mt: 1.25, mr: 1.25 }}
+            >
+              My app
+            </Typography>
+
+            <Box sx={{ ml: "auto" }}>
+              <ColourSchemeButton />
+            </Box>
+          </Toolbar>
+        </AppBar>
+
+        <NavigationLayout
+          navigation={navigation}
+          sidebarOpen={sidebarOpen}
+          setSidebarOpen={setSidebarOpen}
+          secondaryNav={{ title: "Setup", groups: setupGroups }}
+          secondaryNavOpen={secondaryNavOpen}
+          setSecondaryNavOpen={setSecondaryNavOpen}
+        >
+          <Typography variant="h5">Main content here</Typography>
+        </NavigationLayout>
+      </Box>
+    );
+  },
+  parameters: {
+    docs: {
+      description: {
+        story:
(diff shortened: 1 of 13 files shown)


src/components/navigation/SecondaryNav.tsx
```diff
@@ -0,0 +1,421 @@
...
+
+const SECONDARY_NAV_WIDTH = 256; // matches SidebarNav's open-state baseline width
+
+type SecondaryNavGroup = {
+  /** Rendered as an overline ListSubheader when present; omit for an ungrouped list. */
+  subheader?: string;
+  items: SecondaryNavItemDefinition[];
+};
+
+type SecondaryNavChildItemDefinition = {
```
douglaswinter: This should be consolidated with the related props used for SidebarNav

src/components/navigation/SecondaryNav.tsx
```diff
@@ -0,0 +1,421 @@
...
+  id: string;
```
douglaswinter: Do we need this, if it's just for setting the react key? In SidebarNav I just use a generated index

src/components/navigation/SecondaryNav.tsx
```diff
@@ -0,0 +1,421 @@
...
+  type ReactNode,
+} from "react";
+import useMediaQuery from "@mui/material/useMediaQuery";
+import ArrowBackIcon from "@mui/icons-material/ArrowBack";
+import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
+import SearchIcon from "@mui/icons-material/Search";
+import { drawerTransition } from "./SidebarNav";
+import type { LinkProps } from "./types";
+
+const SECONDARY_NAV_WIDTH = 256; // matches SidebarNav's open-state baseline width
```
douglaswinter: Should have a shared definition used in both components. Note that in SidebarNav I need +1 because of a border

src/components/navigation/SecondaryNav.tsx
```diff
@@ -0,0 +1,421 @@
...
+
+type SecondaryNavChildItemDefinition = {
+  id: string;
+  label: string;
+  icon?: ReactNode;
+  linkProps?: LinkProps;
+  selected?: boolean;
+};
+
+type SecondaryNavItemDefinition = SecondaryNavChildItemDefinition & {
```
douglaswinter: The names SecondaryNavChildItemDefinition and SecondaryNavItemDefinition could be clearer... How about NavItem and NavItemWithChildren?

zoharma:
@douglaswinter Have updated a few things:

- Fixed `WithAppBar` story: selection state was hardcoded to "Acquisition" and now fixed.
- Split `SecondaryNav` content from its responsive layout (added to `NavigationLayout`), making it reusable (such as in modals)
- Fixed mobile navigation so selecting a secondary item closes both sidebars and shows the main content.
- Updated tests and stories.
- Fixed how `NavigationLayout` looks and works on the docs page

src/components/navigation/NavigationLayout.tsx
```diff
@@ -0,0 +1,161 @@
...
+            }}
+          >
+            <Toolbar />
+            <SecondaryNavContent
+              {...props.secondaryNav}
+              onBack={() => props.setSecondaryNavOpen(false)}
+            />
+          </Drawer>
+        ))}
+      <Box component="main" sx={{ flexGrow: 1, minWidth: 0 }}>
```
douglaswinter: I think this component should do one thing only: the composition of SidebarNav with SecondaryNav. In that sense, it should not take children which are here then rendered as siblings - that is someone else's concern. I think this component should therefore end in line 119.

Another component, further down the line, could provide the unified layout of an app, including appbar, etc, and that one should take children so it can effectively do something like
```
<Appbar />
<Box>
  <SidebarNav />
  <main />
</Box>
```

src/components/navigation/NavigationLayout.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Composes SidebarNav with SecondaryNavContent, owning all of the responsive
+ * behaviour between them: on mobile only one temporary drawer can be visible
+ * at a time, so drilling into the secondary content implicitly hides the
+ * primary sidebar, and its back affordance is a pure consequence of flipping
+ * `secondaryNavOpen` back to false. On desktop both are shown side by side,
+ * the secondary content in a plain panel rather than a drawer.
+ */
+function NavigationLayout(props: NavigationLayoutProps) {
```
douglaswinter: I mentioned this to you in person: I think what this component does should be the behaviour for SidebarNav anyway, as soon as the navigation tree goes beyond a first level. Apps should use SidebarNav, which depending on the tree renders either PrimaryNav (currently called SidebarNav), or PrimaryNav and SecondaryNav.

src/components/navigation/SecondaryNav.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+      <MemoryRouter>
+        <Story />
+      </MemoryRouter>
+    ),
+  ],
+  tags: ["autodocs"],
+  parameters: {
+    docs: {
+      description: {
+        component: `The content of an optional contextual navigation panel that sits next to SidebarNav: a header (title/search/back) plus a grouped, optionally-expandable list. Use NavigationLayout to get the responsive mobile drill-down / desktop side-by-side drawer-or-panel behaviour around it.`,
```
douglaswinter: Should not say it sits next to SidebarNav here. If it can exist as a standalone component we need to clearly state what it is (and probably rename it).

src/components/navigation/SecondaryNav.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+      {
+        id: "setup",
+        label: "Setup",
+        linkProps: { to: "/1", component: NavLink },
+      },
+      {
+        id: "acquisition",
+        label: "Acquisition",
+        linkProps: { to: "/2", component: NavLink },
+        selected: true,
```
douglaswinter: Remove this property

src/components/navigation/NavigationLayout.stories.tsx
douglaswinter, line 1: There is an unexpected scroll bar associated with SecondaryNav in these stories

src/components/navigation/SecondaryNav.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+  });
+
+  it("dense can be turned off for taller rows", () => {
+    renderSecondaryNavContent({ dense: false });
+    expect(screen.getByRole("link", { name: "Setup" })).not.toHaveClass(
+      "MuiListItemButton-dense",
+    );
+  });
+
+  it("never renders a Drawer itself - it has no responsive presentation of its own", () => {
```
douglaswinter: I would remove this test - I don't think it tells us much.

src/components/navigation/SecondaryNav.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+    renderSecondaryNavContent();
+    expect(document.querySelector(".MuiDrawer-root")).not.toBeInTheDocument();
+  });
+
+  it("does not render a header when no header props are provided", () => {
+    renderSecondaryNavContent();
+    expect(screen.queryByRole("searchbox")).not.toBeInTheDocument();
```
douglaswinter: The problem here is that in the next test, which has a search box, screen.queryByRole("searchbox") returns null.

src/components/navigation/SecondaryNav.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+    const input = screen.getByPlaceholderText("Search");
+    await user.type(input, "a");
+
+    expect(onChange).toHaveBeenCalledWith("a");
+    // groups are rendered unfiltered regardless of search value
+    expect(screen.getByRole("link", { name: "Setup" })).toBeVisible();
+  });
+
+  it("renders a back button only when onBack is provided", async () => {
```
douglaswinter: This is testing that it "renders a back button when onBack is provided". You can include a similar test without passing onBack to prove that the Back button is not rendered.

src/components/navigation/SecondaryNav.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+    expect(screen.queryByText("Analysis A")).not.toBeInTheDocument();
+
+    // Clicking the label text, not the chevron IconButton - regression test
+    // for the chevron previously being nested inside the row's own button.
+    await user.click(screen.getByText("Analysis"));
+
+    expect(screen.getByText("Analysis A")).toBeVisible();
+  });
+
+  it("a row with both linkProps and children navigates and toggles together on label click", async () => {
```
douglaswinter: Can you rewrite these titles to make them read nicely with the 'it' function name? e.g. it toggles this and that together on label click

src/components/navigation/SecondaryNav.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+    const link = screen.getByRole("link", { name: "Expandable link" });
+    expect(link).toHaveAttribute("href", "https://www.example.com");
+
+    expect(screen.queryByText("Child")).not.toBeInTheDocument();
+
+    await user.click(link);
+    expect(screen.getByText("Child")).toBeVisible();
+  });
+
+  it("a row with both linkProps and children can also be toggled via the chevron alone", async () => {
```
douglaswinter: and this one

src/components/navigation/SecondaryNav.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+    });
+
+    expect(screen.getByText("Analysis A")).toBeVisible();
+  });
+
+  // A consumer (e.g. NavigationLayout's mobile drawer) may close itself on
+  // any click that bubbles out - these confirm which rows let that happen
+  // and which stop it, independent of any particular consumer.
+  describe("click propagation", () => {
+    it("a plain link row's click bubbles up to an ancestor", async () => {
```
douglaswinter: also these titles. make them all sensible 'it' statements about click propagation

src/components/navigation/NavigationLayout.test.tsx
```diff
@@ -0,0 +1,244 @@
...
+        initialSidebarOpen: true,
+        initialSecondaryNavOpen: true,
+      });
+
+      expect(screen.getByRole("link", { name: "Setup" })).toBeVisible();
+      expect(screen.getByRole("heading", { name: "Secondary" })).toBeVisible();
+      expect(screen.getByRole("link", { name: "Detail" })).toBeVisible();
+    });
+
+    it("renders only SidebarNav as a Drawer - the secondary panel is a plain flex sibling, not a second fixed-position Drawer", () => {
```
douglaswinter: Remove this test - the implementation of the children is the responsibility of the children

src/components/navigation/NavigationLayout.test.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+    it("the device back action drills back to the sidebar instead of leaving the page", async () => {
+      renderHarness({
+        initialSidebarOpen: true,
+        initialSecondaryNavOpen: true,
+      });
+
+      expect(screen.queryByText("Setup")).not.toBeInTheDocument();
+
+      window.dispatchEvent(new PopStateEvent("popstate"));
```
douglaswinter: What are we simulating here? This is just a notification I believe

src/components/navigation/SecondaryNav.tsx
```diff
@@ -0,0 +1,344 @@
...
+  /** One level only - children cannot themselves expand. */
+  children?: SecondaryNavChildItemDefinition[];
+  /** Initial Collapse state for this item; uncontrolled thereafter. */
+  defaultExpanded?: boolean;
+};
+
+type SecondaryNavContentProps = {
+  title?: string;
+
+  search?: {
```
douglaswinter: Following the trend in sci-react, shouldn't this be replaced with a slot system? A search box could become its own component

douglaswinter (changes requested):
It looks good, but I think it could do with some refactoring.

src/components/navigation/NavigationLayout.tsx
zoharma: SidebarNav is now the primary drawer with an optional secondary panel itself (logic lives inside)

zoharma:
@douglaswinter Went through this today:

In two commits here are the changes:

**Address SidebarNav review feedback:**
- Move responsive primary/secondary panel coordination into `SidebarNav`
- Rename the secondary panel component to `SubNav`
- Consolidate shared nav item types and use generated keys instead of `id`
- Replace the `search` prop with `searchSlot`
- Improve `SidebarNav` story coverage and clean up story canvases
- Update test assertions
- Added and fixed story docs for `SidebarNav` and `SubNav`

**Additional changes and fixes:
Add SubNav slots and make AppBar spacer opt-in**
- Replace `SubNav`'s `searchSlot` with `beforeNavSlot`, `afterNavSlot`, and `footerSlot`
- Add `hasAppBar` to `SidebarNav`, defaulting to `false`, so the Toolbar spacer is only rendered when needed
- Align subheaders, expandable-item children, and slot content with nav item text
- Replace the expand/collapse chevron with Lucide's `ChevronDown`
- Expand `SidebarNav` and `SubNav` story docs with usage guidance

PR #267, opened 2026-07-23
Updated docs theme colours and add reference to highlight and tertiary

zoharma:
Correct MUI role calls, regroup and introduce brand colours

- Corrected mislabeled MUI palette calls in the Neutral colours reference (e.g. `onSurfaceMuted` → `text.muted`) to match the real roles in DiamondDSTheme.ts.
- Split "Intent colours" into two groups: functional MUI-exposed roles (primary, secondary, info, danger, warning, success) and a new "Core brand colours" section (brand, tertiary, highlight). _Only brand is still an actual MUI palette role._ 
- Added a contrastText row to each intent/brand family, previously undocumented.

p.s. We can name Tertiary to something else like **Brand Accent** if we thing it would be helpful

### DIFF
--- a/src/storybook/foundation/1.1. theme-colours.mdx
+++ b/src/storybook/foundation/1.1. theme-colours.mdx
@@ -21,22 +21,97 @@ This page is a token reference. If you're deciding which colour role to use, rea
  <section className="ds-mode-panel" data-mode="light">
   <section>
    <h3>Intent colours</h3>
+   <p>
+    Exposed as MUI palette roles (e.g. <code>theme.palette.primary.main</code>
+    ).
+   </p>
    <div className="ds-intent-grid">
     {[
      ["primary", "Primary", false],
      ["secondary", "Secondary", false],
-     ["tertiary", "Tertiary", false],
-     ["brand", "Brand", true],
+     ["info", "Info", false],
      ["danger", "Danger", false],
      ["warning", "Warning", false],
      ["success", "Success", false],
-     ["info", "Info", false],
     ].map(([key, label, hasFixed]) => (
      <div className="ds-colour-family" key={key}>
       <div className="ds-colour-stack">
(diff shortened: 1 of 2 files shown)


src/storybook/foundation/1.1. theme-colours.mdx
```diff
@@ -241,60 +393,60 @@ This page is a token reference. If you're deciding which colour role to use, rea
...
      </div>
      <div className="ds-colour-row ds-bg-on-surface-muted ds-fg-surface">
       <span>On Surface Muted</span>
-      <code>onSurfaceMuted</code>
+      <code>text.muted</code>
      </div>
      <div className="ds-colour-row ds-bg-action-disabled ds-fg-surface">
       <span>Action Disabled</span>
-      <code>action.disabled</code>
+      <code>--ds-action-disabled</code>
```
gfrn: Seems like all other tags are following the MUI naming pattern (`x.y`) instead of the CSS var name, was this intended?

zoharma: Your comment was exposing a bigger issue with token confusion, which is now fixed.

zoharma:
Followng @gfrn's comment, I found a few discrepancies that were fixed as part of this PR:

- `--ds-action-disabled` was only used for disabled Checkbox/Radio icons, so rename it to `--ds-control-disabled` to better reflect its scope.
- Document `--ds-on-surface-disabled` as backing both `text.disabled` and `action.disabled`
- Document `--ds-surface-disabled` as `action.disabledBackground`
- Fix the incorrect “On Surface Disabled” swatch
- Show actual MUI palette paths on intent/brand swatches (for example, `primary.main` and `error.main`)
- Rename “{role} Contrast Text” to “On {role}” for consistency

gfrn:
It seems to work OK for me when deployed locally, and I can't find any more discrepancies, so if you're happy with it, looks good to me

PR #268, opened 2026-08-03
Rework Navbar surface options and fix logo tone matching

zoharma:
- Keep default Navbar surface to `brand` (solid), but with updated label; document `brand-fixed`/`brand-fixedDim` as the alternative for a persistent Diamond identity colour that stays constant across modes
- Add a bottom border to any surface that resolves to a neutral background  (surface/paper/background, and `base` variant on a semantic intent) so it doesn't blend into the content below; saturated surfaces (solid brand/primary/secondary, brand-fixed/brand-fixedDim) provide their own separation
- Fix Navbar logo tone: force the light-on-dark logo only for surfaces that stay saturated in both modes, follow the page mode otherwise (previously hardcoded to "inverse" for every non-brand surface, which put the wrong logo on the new surface base/container options); same fix applied to custom (non-theme) logos
- Document the surface/variant matrix on Bar's props and in both components' Storybook docs
- Updated minimum Navbar to `48px` and added a regression test to `Bar.test.tsx` asserting that change and checking for future ones. Removed regression test to `Footer.test.tsx`.

### DIFF
--- a/src/components/navigation/Navbar.tsx
+++ b/src/components/navigation/Navbar.tsx
@@ -112,7 +112,19 @@ type NavbarProps = BarSlotsProps & {
 };
 
 /**
- * Basic navigation bar. Can be used with `NavLinks` and `NavLink` to display a responsive list of links. Brand surface by default.
+ * Basic navigation bar. Can be used with `NavLinks` and `NavLink` to display a responsive list of links.
+ *
+ * Defaults to the `brand` surface (solid), which adapts with light/dark mode like the rest of
+ * the theme. Alternatives include `brand-fixed`/`brand-fixedDim` for a persistent Diamond
+ * identity colour that stays constant across modes, and the neutral `surface` in `base` or
+ * `container` variant. Most Diamond apps should probably reach for `brand-fixed` or
+ * `surface`/`base` rather than the default — a constant Diamond identity colour, or a plain
+ * neutral bar, tend to suit real product chrome better than a navbar that shifts with the mode.
+ *
+ * `brand-fixed`/`brand-fixedDim` (any variant) and `solid` variant on `brand`/`primary`/`secondary`
+ * resolve to a fully saturated colour that provides its own separation from the page. Every other
+ * combination (including `container`, and `base` on a semantic intent, which falls back to a
+ * plain neutral background) picks up a bottom border so it doesn't blend into the content below.
  */
 const Navbar = ({
   surface = "brand",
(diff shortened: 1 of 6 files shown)


zoharma:
**Logo tone bug fix:** the logo was hardcoded to `tone="inverse"` for every surface except `brand/brand-fixed`. That was invisible while those were the only documented options, but it puts the wrong-toned logo on the new `surface base/surface container` options (and on custom, non-theme logos). Logo tone now follows the same saturated-vs-neutral classification as the border.

src/components/navigation/Navbar.tsx
```diff
@@ -112,7 +112,19 @@ type NavbarProps = BarSlotsProps & {
 };
 
 /**
- * Basic navigation bar. Can be used with `NavLinks` and `NavLink` to display a responsive list of links. Brand surface by default.
+ * Basic navigation bar. Can be used with `NavLinks` and `NavLink` to display a responsive list of links.
+ *
+ * Defaults to the `brand` surface (solid), which adapts with light/dark mode like the rest of
+ * the theme. Alternatives include `brand-fixed`/`brand-fixedDim` for a persistent Diamond
+ * identity colour that stays constant across modes, and the neutral `surface` in `base` or
+ * `container` variant. Most Diamond apps should probably reach for `brand-fixed` or
```
gfrn: If I understood correctly, this calls for a constant surface colour, rather than one that updates dynamically with the colour mode.

Maybe we need some explicit guidance on this? I guess nowadays most websites opt for a neutral colour or the background colour with a divider (like BBC, Youtube, etc.), so we could probably mention that matching the background (or a neutral colour) would be preferrable to a fixed colour (like blue)

zoharma: I had it in stories `docs` page, but updated it slightly, though I'm, not giving direct order:
> Most Diamond apps should use `brand-fixed` or `surface/base` rather than the default. Having a constant identity colour or plain neutral bar suits real product chrome better than one that shifts with the mode.

Also, add some of this to simplified comments on `Navbar.tsx`.

We could make it more explicit, and define a ruleset.

zoharma: Actually I'd like to improve it further: "A constant identity colour or neutral header can work better for persistent product chrome, providing a stable visual anchor across light and dark modes, even when its tone shifts to suit the colour scheme."

zoharma: This is done @gfrn

gfrn: Yeah, I think that is far clearer and provides some real solid reasoning! Looks good to me

PR #221, opened 2026-06-16
New component: Initial SidebarNav implementation

douglaswinter:
Closes #210

PR #270, opened 2026-08-04
Add afterNavSlot and footerSlot to SidebarNav

zoharma:
This is built on top of #221 

- Lets callers inject content after the nav items (inside the scrollable area) or pinned to the bottom of the drawer, e.g. secondary links or a settings/user menu
- Also fixes long labels like "Data Browse" wrapping to two lines and inflating row height while the sidebar is collapsed.

### DIFF
--- a/src/components/navigation/SidebarNav.stories.tsx
+++ b/src/components/navigation/SidebarNav.stories.tsx
@@ -1,0 +1,139 @@
+import {
+  Abc,
+  ArrowForward,
+  CorporateFare,
+  GraphicEq,
+  Menu,
+} from "@mui/icons-material";
+import { SidebarNav } from "./SidebarNav";
+import { Meta, StoryObj } from "@storybook/react";
+import React from "react";
+import {
+  AppBar,
+  Box,
+  Divider,
+  IconButton,
+  Toolbar,
+  Typography,
+} from "../MUI/MuiWrapped";
+import { Theme } from "@mui/material/styles";
+import { Logo } from "../controls/Logo";
+import { ColourSchemeButton } from "../controls/ColourSchemeButton";
+
+const meta: Meta<typeof SidebarNav> = {
+  title: "Components/Navigation/SidebarNav",
+  component: SidebarNav,
+  tags: ["autodocs"],
+  parameters: {
+    docs: {
+      pages: {},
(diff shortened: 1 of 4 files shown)


src/components/navigation/SidebarNav.stories.tsx
```diff
@@ -190,21 +207,120 @@ const groupedNavigation = [
...
+    const [open, setOpen] = React.useState(true);
+    return (
+      <Box sx={{ display: "flex" }}>
+        <SidebarNav
+          navigation={groupedNavigation}
+          open={open}
+          setOpen={setOpen}
+          afterNavSlot={
+            <SlotOutline label="afterNavSlot" open={open}>
+              <ListItem disablePadding>
```
gfrn: Really minor nitpick, this `ListItem` is currently orphaned, it does not sit under a `List` component, although it doesn't really degrade anything in an obvious way to me

src/components/navigation/SidebarNav.tsx
```diff
@@ -87,13 +91,14 @@ function PermanentDrawer(props: NavProps) {
         transition: (theme: Theme) => drawerTransition(theme, props.open),
         [`& .MuiDrawer-paper`]: {
           width: width,
+          height: "100vh",
```
gfrn:
```suggestion
```

I'm not entirely sure this is necessary, as a matter of fact, it seems to be ignored

Image

`100%` is also the safer option, because since this is `fixed`, `100vh` can behave weirdly on mobile devices (https://stackoverflow.com/questions/37112218/css3-100vh-not-constant-in-mobile-browser), so we can probably remove this as 100% is already set by `MuiPaper-root`

(Unless there is a reason to have this over here; I removed it and it doesn't seem to have done anything)

gfrn:
Looks good to me! I left two small comments, but they aren't bugs or breaking anything, so I'm happy to approve once these two are resolved

src/components/navigation/SidebarNav.stories.tsx
zoharma: The `list` was added at higher in the tree, but it isn't very clean, so updated.

zoharma:
Thank you @gfrn this is now updated

PR #271, opened 2026-08-05
Add increment/decrement spinner buttons to NumberInput

### LINKED ISSUE #273: Change Proposal: Increment/Decrement Spinner Buttons for `NumberInput`
## What is being proposed?

Add optional up/down spinner buttons to `NumberInput` so users can nudge a numeric value by a configurable step, instead of only being able to type a value directly.

Image

## Why is this needed?

`NumberInput` currently only accepts typed input. For small, discrete adjustments (e.g. incrementing an integer count), typing is slower and more error-prone than a click-to-step control, which is a common affordance for numeric fields. There was previously no way to adjust the value without full keyboard entry.

## What will change?

- Behaviour: `NumberInput` can now render a vertical up/down `ButtonGroup` inside the field as an end adornment. Clicking increases/decreases the current value by step, clamped to `minValue`/`maxValue`. Buttons disable automatically at the limits.

- New props:
  - `step?: number`: the increment/decrement amount, defaults to `1`.
  - `spinnerButtons?: "always" | "onHover" | "never"`: controls spinner visibility. Defaults to `"onHover"` for all number modes except `"scientific"`, which defaults to `"never"` (since stepping isn't a natural operation on scientific-notation values).
  - No new component: this extends the existing `NumberInput`.

## Interface changes (if any)

Additive/optional only — existing usages of NumberInput are unaffected.

```ts
<NumberInput
  label="Quantity"
  numberMode="natural"
  step={1}
  spinnerButtons="always" // "always" | "onHover" | "never"
/>
```

## Breaking change?

- [ ] Yes
- [X] No

## Next steps
A maintainer will review this issue.
If accepted, it will be marked as `accepted` and a PR may then be opened.

zoharma:
Adds a vertical up/down ButtonGroup and a step prop to `NumberInput` so values can be nudged without typing. Visibility is configurable via spinnerButtons ("always" | "onHover" | "never"), defaulting to on-hover for every mode except scientific.

image

### DIFF
--- a/src/components/controls/NumberInput.stories.tsx
+++ b/src/components/controls/NumberInput.stories.tsx
@@ -13,6 +13,21 @@ const meta: Meta<typeof NumberInput> = {
       },
     },
   },
+  argTypes: {
+    spinnerButtons: {
+      control: "select",
+      options: ["always", "onHover", "never"],
+      description:
+        "Controls visibility of the increment/decrement buttons. Defaults to 'onHover', except for numberMode='scientific' which defaults to 'never'.",
+      table: { category: "Behaviour" },
+    },
+    step: {
+      control: "number",
+      description:
+        "Amount the increment/decrement buttons change the value by.",
+      table: { category: "Behaviour" },
+    },
+  },
 };
 
 export default meta;
--- a/src/components/controls/NumberInput.test.tsx
+++ b/src/components/controls/NumberInput.test.tsx
@@ -158,4 +158,18 @@ describe("NumberInput", () => {
     fireEvent.change(numberInput, { target: { value: "5e5" } });
     expect(screen.queryByText("Invalid input")).not.toBeInTheDocument();
   });
+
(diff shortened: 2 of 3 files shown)


EmsArnold:
I've managed to produce a rounding error when you have a floating point number (started at 4.1, and moves to something like 3.100000000005). This happens inconsistently, and doesn't seem to be tied to the number of decimal places specified. I'm not sure of the correct solution to deal with this.

zoharma:
@EmsArnold Added a fix and a test for the floating point drift

EmsArnold:
Are the up/down arrows meant to be keyboard controllable? I can't seem to activate them by any combination of up/down, left/right, or tab+return

zoharma:
@EmsArnold If you used the tab to move across (focus on page first), it will select inside the field then the first of the two buttons, then space will work to increment

EmsArnold (approved):
LGTM, functions as intended and a test to ensure incrementing works.

